PR #5, opened 2021-01-18
OPSEXP-700: Fix helm related pre-commit checks

gavincornwell:
OPSEXP-700 Pre commit hooks are cached on install or first run which can cause undesirable results when trying to update the hooks themselves therefore removed caching from build and manually delete the folder before running checks.

Furthermore, the "entry" property in .pre-commit-hooks.yaml needs to be a single quoted string.

Finally, the kubebug checks were running against a cluster and not the provided input files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 repos:
   - repo: https://github.com/Alfresco/alfresco-build-tools
-    rev: master
+    rev: v1.0.0
     hooks:
       - id: travis-yml-lint
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -49,7 +49,7 @@
   pass_filenames: false
 - id: checkov-helm
   name: Checkov - Helm check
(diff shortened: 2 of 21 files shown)


mteodori:
@gavincornwell is it correct to remove the cache for pre-commit or maybe enough not to use the branch as ref in .pre-commit.config.yml but always the short sha as in `git rev-parse --short HEAD`?

gavincornwell:
@mteodori No, it's not enough, changing the ref to *anything* has no effect once the pre-commit hooks have been run once, this silent failure to fetch the latest code from the remote repo is what caused to me spend ages on this problem!

As this will only run once per build I'm not sure having it cached is that important to be honest.

mteodori:
> @mteodori No, it's not enough, changing the ref to _anything_ has no effect once the pre-commit hooks have been run once, this silent failure to fetch the latest code from the remote repo is what caused to me spend ages on this problem!
> 
> As this will only run once per build I'm not sure having it cached is that important to be honest.

maybe we have to measure, but on the same PR I would avoid to download the same hooks multiple times
I am a bit skeptic about the lack of effect of rev as I would see the same behaviour locally and also other Travis builds I have seen the changes applied

gavincornwell:
Ah, actually I see what you mean, if you changed the rev *every* time you made a change in the alfresco-build-tools I can see how that might work, I haven't tried that myself.

I will revert the cache changes.

mteodori (approved):

PR #18, opened 2022-02-22
OPSEXP-1266: checkov2 makes sense for lint stages only with python 3.8+

alxgomz:
Forces checkov2 installation to happen with  python3.9 and avoid running during others stages thatn linters.

### DIFF
--- a/.travis.checkov_install.yml
+++ b/.travis.checkov_install.yml
@@ -4,9 +4,9 @@ env:
     - CHECKOV_VERSION=${CHECKOV_VERSION:-2.0.795}
 jobs:
   include:
-    - name: checkov
+    - name: pre-commit
       stage: lint
       language: python
       python: 3.9
-      install:
+      before_install:
         - pip install checkov==$CHECKOV_VERSION
--- a/.travis.checkov_install.yml
+++ b/.travis.checkov_install.yml
@@ -2,7 +2,7 @@ language: python
 env:
   global:
     - CHECKOV_VERSION=${CHECKOV_VERSION:-2.0.795}
-jobes:
+jobs:
   include:
     - name: checkov
       stage: lint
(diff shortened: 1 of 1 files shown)


.travis.checkov_install.yml
```diff
@@ -2,5 +2,11 @@ language: python
 env:
   global:
     - CHECKOV_VERSION=${CHECKOV_VERSION:-2.0.795}
+jobs:
+  include:
+    - name: checkov
+      stage: lint
+      language: python
+      python: 3.9
```
gionn, line 10: not sure if it's correct to force a new job here, I see that most things are expected to be handled by pre-commit (like I did in https://github.com/Alfresco/alfresco-ansible-deployment/pull/259/files) since a dedicated job for fast-executing linters is quite a resource waste

alxgomz: Yes good point. I will favor pre-commit over the other ones in the repo I'm calling from. But this still needs to be fixed here as we have no guarantee anyone using it will have pre-commit or want to favor it over these linting jobs, neither are we ure they'd force python 3.9 on their side

alxgomz: This is just an installation job so pre-commit have all the necessary tools to run

EquinoxBlack (approved):

.travis.checkov_install.yml
mteodori: can we test this works when combined with the pre-commit snippet? I agree with @gionn we're deviating from the general behaviour of all the install snippets to installing a tool as an extra before_install on the root so that it can be used anywhere

alxgomz: Indeed, that breaks the pre-commit import. I could set the python version on the root of the imported config but I feel like it's a lot more intrusive and also I don't see the point of having checkov in stages other than linter...

Mieszko96 (approved):

gionn (approved):

PR #29, opened 2022-03-24
OPSEXP-1221: Create checkov action

Mieszko96:


.github/actions/setup-checkov/action.yml
```diff
@@ -0,0 +1,16 @@
...
+    default: 2.0.930
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        CHECKOV_VERSION=${{ inputs.version }}
+        pip install checkov==$CHECKOV_VERSION
+      shell: bash
+      env:
+        CHECKOV_VERSION: ${{ inputs.version }}
```
gionn, line 16:
```suggestion
        pip install checkov==${{ inputs.version }}
      shell: bash
```

should be enough

gionn (changes requested):

gionn (commented):
try to add it to https://github.com/Alfresco/alfresco-build-tools/blob/master/.github/workflows/test.yml in order to test it

gionn (approved):

mteodori:
this should be reverted there is already an official checkov action https://github.com/bridgecrewio/checkov-action

PR #30, opened 2022-03-24
OPSEXP-1221: Add github action awscli

Mieszko96:


.github/actions/setup-aws-cli/action.yml
```diff
@@ -0,0 +1,16 @@
...
+inputs:
+  version:
+    description: 'Version of awscli'
+    required: false
+    default: 2.4.25
+runs:
+  using: "composite"
+  steps:
+    - run: | 
+        AWSCLI_VERSION=${{ inputs.version }}
```
gionn, line 12: prefer using `env` section to provide variables for a run

here you don't really to set a variable, just use ${{ inputs.version }} in curl command

gionn (changes requested):

gionn (changes requested):
lol there is already awscli in ubuntu runners on github actions, see https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md

PR #36, opened 2022-04-01
Document replacement action for auto cancel builds

gionn:


Mieszko96 (approved):

mteodori:
this choice of action is not in sync with the one used in activiti community, unless there are reasons we should stick to the former choice

https://github.com/Activiti/activiti-cloud/blob/develop/.github/workflows/main.yml#L31-L34

PR #37, opened 2022-04-01
Document usage of docker build action for buildx support

gionn:


Mieszko96 (approved):

mteodori:
this is not the correct action to suggest, this one installs the docker buildx plugin and lets scripts decide how to use it https://github.com/docker/setup-buildx-action

PR #38, opened 2022-04-06
feat: AAE-8050 create actions to handle helm build

erdemedeiros:
- update action `calculate-next-internal-version` to include dependencies
Usage:
```yaml
 	  - id: next-release
        name: Calculate next internal release
        uses: Alfresco/alfresco-build-tools/.github/actions/calculate-next-internal-version@ref
        with:
          next-version: 7.3.0

      - name: Update VERSION file
        run: echo ${{steps.next-release.outputs.next-prerelease}} > VERSION
```
- create action `configure-git-for-commit`
Usage:
```yaml
      - name: Configure git user for commit
        uses: Alfresco/alfresco-build-tools/.github/actions/configure-git-for-commit@ref
        with:
          username: ${{ secrets.BOT_GITHUB_USERNAME }}
```
- create action `run-pre-commit-for-helm` including the necessary dependencies to run pre-commit for helm charts
Usage:
```yaml
    - uses: Alfresco/alfresco-build-tools/.github/actions/run-pre-commit-for-helm@ref
```


.github/actions/calculate-next-internal-version/action.yml
```diff
@@ -11,6 +11,16 @@ outputs:
...
   steps:
+    - uses: actions/setup-python@v2
+    - id: get-current-action-branch
+      run: ${{ github.action_path }}/../shared-scripts/get-current-action-branch.sh
+      shell: bash
+    - uses: actions/checkout@v3
+      with:
+        repository: Alfresco/alfresco-build-tools
+        ref: ${{steps.get-current-action-branch.outputs.action-branch}}
+        path: build-tools
```
erdemedeiros, line 22: This is the way I found to reference another Action inside this repository without hard coding the name of the branch/tag: 
1. `get-current-action-branch.sh` is resolving the current branch of this action. Note: `github.ref` cannot be used because it's resolving the branch of the workflow and not one of `build-tools`. 
2. the checkout action is cloning again build tools using the resolved branch

Is there a better way of doing it?

andrea-ligios (approved):
LGTM!

.github/actions/calculate-next-internal-version/action.yml
gionn: I was struggling with that too and this looks a clever workaround, maybe you can move this steps on a separated action called something like 'setup-alfresco-build-tools' so that can be reused in any other composite action?

The unique downside I can think of that in this way we are enforcing that all actions coming from this repo must be at the same version (but maybe it's what we want to avoid complexity for using different versions for different actions within the same workflow)

.github/actions/calculate-next-internal-version/action.yml
```diff
@@ -11,6 +11,16 @@ outputs:
...
   using: "composite"
   steps:
+    - uses: actions/setup-python@v2
+    - id: get-current-action-branch
+      run: ${{ github.action_path }}/../shared-scripts/get-current-action-branch.sh
+      shell: bash
+    - uses: actions/checkout@v3
+      with:
+        repository: Alfresco/alfresco-build-tools
+        ref: ${{steps.get-current-action-branch.outputs.action-branch}}
```
gionn, line 21: add spaces between curly brackets for readability

.github/actions/calculate-next-internal-version/action.yml
erdemedeiros: Unfortunately, I didn't find a way to make this reusable (other than using the shared script `get-current-action-branch.sh`). If I move it to a separate composite I'll need to reference it using a version (what this workaround is actually avoiding).

Note that not necessarily all the actions coming from this repo must be on the same version: this is only ensuring that a given Action `A` referencing another Action `B` in this repository will use `B` in the same version as `A`. However, an external workfow using `B` independently could reference `B` in a different version than `A`. For instance,

`- uses: Alfresco/alfresco-build-tools/.github/actions/run-pre-commit-for-helm@v3`
`- uses: Alfresco/alfresco-build-tools/.github/actions/setup-helm-docs@v2`

The `setup-helm-docs` used independently in the workfow will be in version `v2`, but the one referenced by `un-pre-commit-for-helm` will use `setup-helm-docs` in version `v3`.

gionn (approved):

.github/actions/configure-git-user/action.yml
```diff
@@ -0,0 +1,15 @@
+name: 'Configure git user'
```
mteodori, line 1: do we need this? there are plenty of similar generic actions in the marketplace like https://github.com/marketplace/actions/configure-git-credentials

.github/actions/run-pre-commit-for-helm/action.yml
```diff
@@ -0,0 +1,17 @@
+name: "Run pre-commit for helm-charts"
```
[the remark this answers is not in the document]
erdemedeiros: The idea was to avoid duplicating calls to the installation of all the libraries that are needed to check pre-commits for helm charts, which I'm assuming will be always the same. But I missed the part that logic now is split between the pre-commit configuration defined in the repo and the tools that should be installed for that (in build tools), so maybe duplicating a couple of lines it's not a big deal after all. I'm removing this.

.github/actions/configure-git-user/action.yml
erdemedeiros: Actually, I think we could completely replace the current propagation mechanism with an action such as this one https://github.com/marketplace/actions/create-pull-request. But I'll do it in a second iteration and only skip this action for now.

gicappa (approved):
LGTM!

PR #44, opened 2022-04-14
Add tip about missing realpath on macosx

gionn:


README.md
```diff
@@ -144,3 +144,19 @@ jobs:
...
+
+### realpath not available under macosx
+
+When running pre-commit locally you may get failures with the following error:
+
+```
+realpath: command not found
+```
+
+This is because macosx lacks support for that, and can be fixed via:
```
gionn, line 158:
```suggestion
This is because macosx lacks support for that, and it can be fixed with:
```

alxgomz (approved):
LGTM

PR #47, opened 2022-04-29
Do not checkout in pre-commit action

gionn:
Initially it looked handy given that pre-commit is often the first step of a job, but unfortunately that will overwrite any previous checkout invocation for which the user may have provided additional parameters.

Required in https://github.com/Alfresco/security-testing-travis/pull/15

### DIFF
--- a/.github/actions/pre-commit/action.yml
+++ b/.github/actions/pre-commit/action.yml
@@ -8,7 +8,6 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/checkout@v3
     - name: Set up Python ${{ inputs.python-version }}
       uses: actions/setup-python@v3
       with:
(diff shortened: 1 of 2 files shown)


Mieszko96 (approved):

mteodori (changes requested):
this pre-commit action doesn't look particularly valuable, I would delete it completely

gionn:
> this pre-commit action doesn't look particularly valuable, I would delete it completely

it's just a starter for now, in the future I think we can use it as an hook to handle pre-commit for multiple repos in a single place

I am going to close this since if some repository requires a particular checkout configuration, is sufficient to move the checkout AFTER executing pre-commit to avoid overwriting it.

mteodori:
Let's delete it, it's a few line of code of reuse that bring more side effect than value thanks

PR #60, opened 2022-05-19
AAE-8518 add utility GH Actions/workflows to build helm projects

erdemedeiros:


.github/actions/calculate-next-internal-version/action.yml
```diff
@@ -12,15 +12,7 @@ runs:
...
-    - id: get-current-action-branch
-      run: ${{ github.action_path }}/../shared-scripts/get-current-action-branch.sh
-      shell: bash
-    - uses: actions/checkout@v3
-      with:
-        repository: Alfresco/alfresco-build-tools
-        ref: ${{ steps.get-current-action-branch.outputs.action-branch }}
-        path: build-tools
-    - uses: ./build-tools/.github/actions/setup-pysemver
+    - uses: Alfresco/alfresco-build-tools/.github/actions/setup-pysemver@v1.4.0
```
erdemedeiros, line 15: Unfortunately, the workaround I used to avoid referencing the version of build-tools inside the project itself is not working in all the scenarios. I.e. it's not working when the cation is invoked by a reusable workflow. The other alternative would systematically use `actions/checkout` to checkout `build-tools` in workflows and then reference the actions using a relative path, but think it will be less intuitive.

gionn: I still don't have any strong opinion on a solution for this problem that is mainly a lack of flexibility on GHA side.

An alternative to your proposal (checkout + relative paths) that is simplest but not DRY at all is **avoid calling an action inside an action**, that also avoid the problem of actions with a lot of parameters where you would need to support overriding all parameters otherwise caller would be stuck with defaults. That would also enable to use different action versions in the same workflow, don't know if it's a feature or not.

we probably need to experiment a bit and then take a decision on this before we start adding more and more custom actions.

erdemedeiros: It's true that overriding parameters may become cumbersome, but at the same time, I'm not comfortable with repeating the same blocks again and again across several repositories.

.github/workflows/helm-publish-new-package-version.yml
```diff
@@ -0,0 +1,86 @@
...
+        required: false
+    outputs:
+      version:
+        description: "The version of the new published pacakage"
+        value: ${{ jobs.publish.outputs.version }}
+
+jobs:
+  publish:
+    runs-on: ubuntu-20.04
+    if: ${{ github.event_name == 'push' }}
```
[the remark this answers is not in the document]
erdemedeiros: A reusable workflow acts as if it was defined inside the caller. For instance, if a reusable workflow uses `actions/checkout`, it will checkout the repository of the caller, not the repository where the reusable workflow is defined. But that's a good point, it's probably better to move this restriction to the caller level, so it's more flexible. I'll do that.

.github/workflows/helm-publish-new-package-version.yml
```diff
@@ -0,0 +1,86 @@
...
+        type: string
+        required: false
+    outputs:
+      version:
+        description: "The version of the new published pacakage"
+        value: ${{ jobs.publish.outputs.version }}
+
+jobs:
+  publish:
+    runs-on: ubuntu-20.04
```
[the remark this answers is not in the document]
erdemedeiros: No, there is no particular reason for that. I'll change it to use latest.

mteodori (commented):
can we just use the standard actions provided by the helm GitHub org and drop our custom process? for helm lint it's already done via pre-commit so we can remove a duplicated run

mteodori (approved):

PR #71, opened 2022-06-07
AAE-8964 add reusable action and workflows

erdemedeiros:
Needed by:
- https://github.com/Alfresco/alfresco-modeling-service/pull/284
- https://github.com/Alfresco/alfresco-process-connector-services/pull/1560
- https://github.com/Alfresco/alfresco-process/pull/913
- https://github.com/Alfresco/alfresco-deployment-service/pull/556

### DIFF
--- a/.github/actions/update-pom-to-next-pre-release/action.yml
+++ b/.github/actions/update-pom-to-next-pre-release/action.yml
@@ -1,6 +1,9 @@
 description: Update pom files to the next pre-release
 name: Update pom to next pre-release
 inputs:
+  maven-cli-opts:
+    description: extra maven properties
+    required: false
   property-to-update:
     description: property to update in additon to the version of the pom file
     required: false
@@ -28,11 +31,12 @@ runs:
     - name: Update pom files to the new version
       shell: bash
       env:
+        MAVEN_CLI_OPTS: ${{ inputs.maven-cli-opts }}
         VERSION: ${{ steps.next-prerelease-resolver.outputs.next-prerelease }}
         PROPERTY_NAME: ${{ inputs.property-to-update }}
       run: |
-        mvn -B versions:set -DnewVersion=$VERSION -DprocessAllModules=true -DgenerateBackupPoms=false
+        mvn -B versions:set -DnewVersion=$VERSION -DprocessAllModules=true -DgenerateBackupPoms=false $MAVEN_CLI_OPTS
         if [ -n "$PROPERTY_NAME" ]
         then
-          mvn versions:set-property -Dproperty=$PROPERTY_NAME -DnewVersion=$VERSION -DgenerateBackupPoms=false
+          mvn versions:set-property -Dproperty=$PROPERTY_NAME -DnewVersion=$VERSION -DgenerateBackupPoms=false $MAVEN_CLI_OPTS
         fi
--- a/.github/actions/update-pom-to-next-pre-release/action.yml
+++ b/.github/actions/update-pom-to-next-pre-release/action.yml
@@ -29,10 +29,10 @@ runs:
       shell: bash
       env:
         VERSION: ${{ steps.next-prerelease-resolver.outputs.next-prerelease }}
-      run: mvn -B versions:set -DnewVersion=$VERSION -DprocessAllModules=true -DgenerateBackupPoms=false
-
-    - name: Update property to the new version
-      shell: bash
-      env:
         PROPERTY_NAME: ${{ inputs.property-to-update }}
-      run: mvn versions:set-property -DnewVersion=$VERSION -DgenerateBackupPoms=false -Dproperty=$PROPERTY_NAME
+      run: |
+        mvn -B versions:set -DnewVersion=$VERSION -DprocessAllModules=true -DgenerateBackupPoms=false
+        if [ -n "$PROPERTY_NAME" ]
+        then
+          mvn versions:set-property -Dproperty=$PROPERTY_NAME -DnewVersion=$VERSION -DgenerateBackupPoms=false
+        fi
--- a/.github/actions/update-pom-to-next-pre-release/action.yml
+++ b/.github/actions/update-pom-to-next-pre-release/action.yml
@@ -1,5 +1,9 @@
 description: Update pom files to the next pre-release
 name: Update pom to next pre-release
+inputs:
+  property-to-update:
+    description: property to update in additon to the version of the pom file
+    required: false
 outputs:
   next-prerelease:
     description: "Next prerelease"
@@ -26,3 +30,9 @@ runs:
       env:
         VERSION: ${{ steps.next-prerelease-resolver.outputs.next-prerelease }}
       run: mvn -B versions:set -DnewVersion=$VERSION -DprocessAllModules=true -DgenerateBackupPoms=false
+
+    - name: Update property to the new version
+      shell: bash
+      env:
+        PROPERTY_NAME: ${{ inputs.property-to-update }}
+      run: mvn versions:set-property -DnewVersion=$VERSION -DgenerateBackupPoms=false -Dproperty=$PROPERTY_NAME
--- a/.github/actions/update-pom-to-next-pre-release/action.yml
+++ b/.github/actions/update-pom-to-next-pre-release/action.yml
@@ -1,0 +1,28 @@
+description: Update pom files to the next pre-release
+name: Update pom to next pre-release
+outputs:
+  next-prerelease:
+    description: "Next prerelease"
+    value: ${{ steps.next-prerelease-resolver.outputs.next-prerelease }}
+runs:
+  using: composite
+  steps:
+    - name: Parse next version from POM
+      id: parse-next-final-version
+      shell: bash
+      run: |
+        NEXT_VERSION=$(yq -p=xml e '.project.version' pom.xml | grep -o "[0-9]*\.[0-9]*.[0-9]*")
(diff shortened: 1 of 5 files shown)


.github/actions/update-pom-to-next-pre-release/action.yml
```diff
@@ -0,0 +1,28 @@
...
+      id: parse-next-final-version
+      shell: bash
+      run: |
+        NEXT_VERSION=$(yq -p=xml e '.project.version' pom.xml | grep -o "[0-9]*\.[0-9]*.[0-9]*")
+        echo "Next final version: $NEXT_VERSION"
+        echo "::set-output name=result::$NEXT_VERSION"
+
+    - id: next-prerelease-resolver
+      name: Calculate next internal release
+      uses: Alfresco/alfresco-build-tools/.github/actions/calculate-next-internal-version@v1.5.0
```
atchertchian, line 20: My understanding (reading the doc) is that you can reference this local action using instead:
```
uses: ./.github/actions/calculate-next-internal-version/action.yml
```
This would allow not having to mention the version (and just use the current one), but I have not tested that it works ok.

.github/workflows/build-and-tag-maven.yml
```diff
@@ -0,0 +1,100 @@
+name: Build and tag Maven project
+
+on:
+  workflow_call:
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Alfresco/alfresco-build-tools/.github/actions/pre-commit@v1.5.0
```
atchertchian, line 10: Same here, maybe you can use the following? (?)
```
- uses: ./.github/actions/pre-commit/action.yml
```

atchertchian, line 10: Same here, maybe you can use the following? (?)
```
- uses: ./.github/actions/pre-commit/action.yml
```

.github/actions/update-pom-to-next-pre-release/action.yml
atchertchian: Also version will be v1.6.0, you will need to tag the repo with this version when the PR has been merged.

atchertchian: Also version will be v1.6.0, you will need to tag the repo with this version when the PR has been merged.

.github/workflows/build-and-tag-maven.yml
```diff
@@ -0,0 +1,100 @@
...
+        uses: actions/setup-java@860f60056505705214d223b91ed7a30f173f6142
+        with:
+          java-version: '11'
+          distribution: 'temurin'
+          cache: 'maven'
+
+      - name: Update pom files to the new version
+        id: update-pom-to-next-version
+        if: ${{ github.event_name == 'push' }}
+        uses: Alfresco/alfresco-build-tools/.github/actions/update-pom-to-next-pre-release@aae-8964-produce-alphas
```
atchertchian, line 28:
```
uses: ./.github/actions/update-pom-to-next-pre-release/action.yml
````
If that change cannot work, the branch name will need to be updated anyway to use the tag instead

Comment (atchertchian) [inline .github/workflows/build-and-tag-maven.yml:28]:
```
uses: ./.github/actions/update-pom-to-next-pre-release/action.yml
````
If that change cannot work, the branch name will need to be updated anyway to use the tag instead

File: .github/workflows/build-and-tag-maven.yml [hunk 4]
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Set version env variable
+        if: ${{ github.event_name == 'push' }}
```
atchertchian, line 31: I wonder if all the checks on `${{ github.event_name == 'push' }}` should be replaced by a check on an input boolean, so that the caller can decide in which cases the push should be done, wdyt?

atchertchian, line 31: I wonder if all the checks on `${{ github.event_name == 'push' }}` should be replaced by a check on an input boolean, so that the caller can decide in which cases the push should be done, wdyt?

.github/workflows/build-and-tag-maven.yml
```diff
@@ -0,0 +1,100 @@
...
+        run: |
+          git tag -a $VERSION -m "Release version $VERSION"
+          git push origin $VERSION
+
+      - name: Slack Notification
+        if: ${{ failure() && github.event_name == 'push' }}
+        uses: rtCamp/action-slack-notify@v2
+        env:
+          SLACK_COLOR: ${{ job.status }}
+          SLACK_WEBHOOK: ${{ secrets.APA_SLACK_WEBHOOK }}
```
atchertchian, line 102: This will definitely need to be extracted as an input parameter. Though I'm not sure the slack notif should be part of the reusable workflow.

atchertchian, line 102: This will definitely need to be extracted as an input parameter. Though I'm not sure the slack notif should be part of the reusable workflow.

.github/actions/update-pom-to-next-pre-release/action.yml
gionn: that's actually the most confusing thing about reusable actions/workflow: referencing an action via a relative/local path would work *only* in this repository. When this action will be used on any other repository that local path don't exists.

Last time we discussed about this I remember that is something that probably GitHub has to solve on their side because with the current status we are pretty limited to use global reference (Alfresco/alfresco-build-tools) and so an explicit ref as version

atchertchian: Indeed Elias explained to me that this does not work as I was expecting :(

erdemedeiros: AFAIK, the only way to have reusable actions / workflows working properly doesn't matter from where they are called, is by referencing them using the repo and version. Unfortunately, we cannot use context expressions to reference the version either.

.github/workflows/build-and-tag-maven.yml
erdemedeiros: yes, while releasing v1.6.0, this will be updated to reference the new tag.

.github/workflows/build-and-tag-maven.yml
atchertchian: this conversation was somehow duplicated (?) --> resolving

.github/workflows/build-and-tag-maven.yml
atchertchian: this conversation was somehow duplicated (?) --> resolving

.github/actions/update-pom-to-next-pre-release/action.yml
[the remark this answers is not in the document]
atchertchian: this conversation was somehow duplicated (?) --> resolving

.github/workflows/build-and-tag-maven.yml
atchertchian: this conversation was somehow duplicated (?) --> resolving

.github/workflows/build-and-tag-maven.yml
erdemedeiros: It makes sense, but I think it's better to do it when we know exactly what we want. For instance, we may need different flags (`pushDocker`, `pushMaven`, `createTag`). For the moment it would only mean an extra parameter to pass while calling the workflow that will have always the same value.

atchertchian: As you prefer :+1:

.github/workflows/build-and-tag-maven.yml
erdemedeiros: I'll move it back to the different repos, as it probably makes sense to add this at the very end of the workflow (after propagations, for instance).

.github/workflows/build-and-tag-maven.yml
```diff
@@ -0,0 +1,107 @@
...
+          fi
+        env:
+          DO_PUSH: ${{ github.event_name == 'push' }}
+      - name: Docker Build (and maybe Push)
+        run: |
+          TAG="${TAG:-$(echo ${GITHUB_HEAD_REF:-$GITHUB_REF_NAME} | sed -e 's/[^-_.[:alnum:]]/_/g')}"
+          sh ./build-and-push-docker-images.sh
+        env:
+          PUSH_OPTION: ${{ steps.define_docker_push.outputs.command }}
+          TAG: ${{env.VERSION}}
```
atchertchian, line 95: nit: this env var should be removed, right?

atchertchian (approved):
:+1: nice work!

.github/workflows/build-and-tag-maven.yml
atchertchian: nevermind sorry (misread the expression)

.github/workflows/build-and-tag-maven.yml
```diff
@@ -0,0 +1,107 @@
...
+        required: false
+    outputs:
+      version:
+        description: "The version of the new tag created by this workflow"
+        value: ${{ jobs.build.outputs.version }}
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Alfresco/alfresco-build-tools/.github/actions/pre-commit@v1.5.0
```
atchertchian, line 18: On second thought, not sure this part should be included in the reusable workflow (as it's conceptually independent of building and tagging)

PR #78, opened 2022-06-30
Ignore not existing file on travis-env-load

gionn:


gionn:
it's not a problem since yq get executed in a subshell so main script exit gracefully anyway

PR #81, opened 2022-07-04
AAE-9367 handle auto merge of propagation PRs

erdemedeiros:
Pull requests created to propagate versions should be automatically approved and merged

### DIFF
--- a/.github/workflows/versions-propagation-auto-merge.yml
+++ b/.github/workflows/versions-propagation-auto-merge.yml
@@ -11,9 +11,11 @@ jobs:
         run: gh pr merge --auto --squash "$PR_URL"
         env:
           PR_URL: ${{github.event.pull_request.html_url}}
+          # this token CANNOT be the default  `GITHUB_TOKEN`, otherwise the merge of the PR will not trigger a build
           GITHUB_TOKEN: ${{secrets.BOT_GITHUB_TOKEN}}
       - name: Auto-approve propagation Pull Request
         run: gh pr review --approve "$PR_URL"
         env:
           PR_URL: ${{github.event.pull_request.html_url}}
+          # This token CANNOT be BOT_GITHUB_TOKEN, because it's linked to alfresco-build user (the user who created the PR)
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
--- a/.github/workflows/versions-propagation-auto-merge.yml
+++ b/.github/workflows/versions-propagation-auto-merge.yml
@@ -3,7 +3,7 @@ on:
   workflow_call:
 
 jobs:
-  dependabot:
+  enable-auto-merge:
     runs-on: ubuntu-latest
     if: ${{ github.actor == 'alfresco-build' && contains( github.event.pull_request.labels.*.name, 'updatebot') }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/versions-propagation-auto-merge.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  enable-auto-merge:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'alfresco-build' && contains( github.event.pull_request.labels.*.name, 'updatebot') }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+      - name: Enable auto-merge for propagation Pull Request
+        run: gh pr merge --auto --squash "$PR_URL"
```
atchertchian, line 14: missing the PR_URL definition in env?

.github/workflows/versions-propagation-auto-merge.yml
```diff
@@ -0,0 +1,21 @@
...
+        env:
+          GITHUB_TOKEN: ${{secrets.BOT_GITHUB_TOKEN}}
+      - name: Auto-approve propagation Pull Request
+        run: gh pr review --approve "$PR_URL"
+        env:
+          PR_URL: ${{github.event.pull_request.html_url}}
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
erdemedeiros, line 21: This token is intentionally different from the one in `line 16`, as we need to have a different user from the one who created the PR to approve it: `github-actions` is approving the PR and `alfresco-build`  is enabling the auto-merge.

atchertchian: :+1: Did not see your comment here before I commented on it below: imho this deserves be be stated as a comment in the code.

.github/workflows/versions-propagation-auto-merge.yml
```diff
@@ -0,0 +1,21 @@
+name: Versions propagation auto-merge
+on:
+  workflow_call:
+
+jobs:
+  enable-auto-merge:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'alfresco-build' && contains( github.event.pull_request.labels.*.name, 'updatebot') }}
```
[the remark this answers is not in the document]
atchertchian: I don't think it's too opinionated: only these PRs should be auto-merged this way. Maybe this can be revisited when there are other use cases that require auto merge but I think it's good like this for now :+1: 
Actually I would even be in favor of adding a specific label on version propagation PRs because the "updatebot" label is quite generic here, but I'm not aware of when the "updatebot" label is set, maybe that matches this only use case (?)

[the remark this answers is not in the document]
erdemedeiros: Yes, the `updatebot` label, is added by the tool `updatebot`  (the one we are using to propagate versions). So it should be all fine, on that end.  The CI is also adding it to PRs that are created using GH client directly.

.github/workflows/versions-propagation-auto-merge.yml
erdemedeiros: Fixed

.github/workflows/versions-propagation-auto-merge.yml
erdemedeiros: I added comments on both tokens, explaining why which one is used.

atchertchian (approved):
:+1:

PR #87, opened 2022-07-08
AAE-8814: add action to auto-merge dependabot minor and patch PRs

atchertchian:


erdemedeiros (approved):
LGTM!

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,37 @@
...
+        env:
+          PR_URL: ${{ github.event.pull_request.html_url }}
+          # this token CANNOT be the default `GITHUB_TOKEN`, otherwise the merge of the PR will not trigger a build
+          GITHUB_TOKEN: ${{ secrets.DEPENDABOT_GITHUB_TOKEN }}
+      - name: Auto-approve patch Dependabot PRs
+        if: ${{ steps.metadata.outputs.update-type == 'version-update:semver-patch' }}
+        run: gh pr review --approve "$PR_URL"
+        env:
+          PR_URL: ${{ github.event.pull_request.html_url }}
+          GITHUB_TOKEN: ${{ secrets.DEPENDABOT_GITHUB_TOKEN }}
```
atchertchian, line 37: @erdemedeiros FYI I made the change here to not use GITHUB_TOKEN (as it will differ from the creator of the PR here anyway)

gionn: secrets needs to be declared in inputs in a reusable workflow (you don't get any error until you try use it - it's always empty even if you are passing on the call)

https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow

gionn: oh it's probably working because of `secrets: inherit` but I would suggest to add them anyway otherwise can be confusing if not using `secrets: inherit` on the caller

README.md
```diff
@@ -353,6 +355,31 @@ Build a maven project and generates the new alpha version for it:
...
 
+### dependabot-auto-merge.yml
+
+Handles automated approval and merge of dependabot PRs, for minor and patch version updates only.
+
+The workflow requires a [dependabot secret](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-encrypted-secrets-for-dependabot) named `DEPENDABOT_GITHUB_TOKEN`, with the `repo > repo:status` and `repo > public_repo` scopes.
+
+```yaml
+  enable-auto-merge:
+    uses: Alfresco/alfresco-build-tools/.github/workflows/dependabot-auto-merge.yml@AAE-8814-improve-dependabot-usage
```
gionn, line 366: check ref - we want to use latest available tag in readme examples

.github/workflows/dependabot-auto-merge.yml
atchertchian: I'm fine following this good practice :+1: 
Now this is not followed by any of the other reusable workflows (also relying on `secrets: inherit` I guess): should this be changed everywhere?
Also I'm not sure how this will behave regarding selection of dependabot secrets (instead of regular secrets): I'd need to test again if that works alright.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,41 @@
...
+        env:
+          PR_URL: ${{ github.event.pull_request.html_url }}
+          # this token CANNOT be the default `GITHUB_TOKEN`, otherwise the merge of the PR will not trigger a build
+          GITHUB_TOKEN: ${{ secrets.token }}
+      - name: Auto-approve patch Dependabot PRs
+        if: ${{ steps.metadata.outputs.update-type == 'version-update:semver-patch' }}
+        run: gh pr review --approve "$PR_URL"
+        env:
+          PR_URL: ${{ github.event.pull_request.html_url }}
+          GITHUB_TOKEN: ${{ secrets.token }}
```
gionn, line 41: can you move these two to the top level `env` given they are the same for both commands - in order to reduce duplication?

gionn (changes requested):

.github/workflows/dependabot-auto-merge.yml
atchertchian: I wanted to keep the specific comment for GITHUB_TOKEN on one of the actions, but ok

gionn (approved):

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.pull_request.user.login == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1.3.3
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
```
erdemedeiros, line 33: Shouldn't this be `${{ env.GITHUB_TOKEN }}`?

atchertchian: My understanding looking at the code is that using the secrets context is the correct way of referencing this token:https://docs.github.com/en/actions/security-guides/automatic-token-authentication

erdemedeiros (approved):

.github/workflows/dependabot-auto-merge.yml
atchertchian: (and this has been tested successfully without the `secrets: inherit` setting from the caller)

.github/workflows/dependabot-auto-merge.yml
atchertchian: (and this has been tested successfully without the secrets: inherit setting from the caller)

PR #88, opened 2022-07-11
remove pre-commit action OPSEXP-1466

himanshu-nigam:


alxgomz:
> Hi, this action is referenced in a lot of repositories, for instance here: https://github.com/Alfresco/alfresco-process/blob/develop/.github/workflows/build.yml#L19. What's the issue with it that requires removal please? Thanks

It's been decided not to create actions if they're simple wrapper and don't bring additionnal value. Here there's no reason why not using the upstream github action. (https://github.com/Alfresco/alfresco-ansible-deployment/pull/294#issuecomment-1117593846)

alxgomz (changes requested):
Please make sure to remove the pre-commit **action** related (not related this repo's pre-commit) doc in the readme file.

alxgomz:
@mteodori , @erdemedeiros ' comment make sense and beyond that, we will find situations where subsequent jobs and steps in Actions rely on the fact repo has already been checked-out and (less likely) python setup...
I would consider that a valid reason to keep that action even though it doesn't provide a huge value

gionn (changes requested):
in the end I will keep it too, not for the current value but for the future improvements potential: we could use it to enforce some pre-commit checks across all the org in a centralized manner.

himanshu-nigam:
Changes not required , keeping Pre-commit for future improvements

PR #93, opened 2022-07-13
AAE-8814: fix issues for dependabot PRs management

atchertchian:


.github/workflows/dependabot-auto-merge.yml
```diff
@@ -31,6 +31,8 @@ jobs:
         uses: dependabot/fetch-metadata@v1.3.3
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
+          # needed to avoid "Dependabot's commit signature is not verified, refusing to proceed" error when synchronizing manually the PR
+          skip-commit-verification: true
```
atchertchian, line 35: actually this could be avoided by asking dependabot to rebase the PR -> not sure it's needed then, wdyt please?

.github/workflows/build-and-tag-maven.yml
```diff
@@ -49,6 +49,7 @@ jobs:
           echo "VERSION=$VERSION" >> $GITHUB_ENV
 
       - name: Login to DockerHub Registry
+        if: ${{ github.event_name == 'push' }}
```
gionn, line 52: I am quite surprised to see this inside a workflow that can be invoked only by `workflow_call`. Does really inherit the event of the caller?

gionn, line 52: additionally, which is the objective here? I am not getting it, please add some context 🙏🏻

.github/workflows/dependabot-auto-merge.yml
gionn: what do you mean with `when synchronizing manually the PR`?

I think it's because dependabot expect that all of its are commits are signed with their key, so for example if you rebase a branch with some dependabot commits in it, will break the signature

atchertchian: I removed this after all because dependabot will not rebase a branch where someones else made some commits.
I mean clicking on the button to merge or rebase the PR when the PR shows the "This branch is out-of-date with the base branch” warning.

.github/workflows/build-and-tag-maven.yml
atchertchian: There are multiple existing checks like this one on this build, and yes, it seems to be inherited.

atchertchian: The objective is to avoid having to log in on DockerHub on PRs, especially dependabot ones because it would require to add corresponding secrets to the dependabot secrets.
I did explain a lot of this on the jira issue, but I can add more comments here if that's better.

atchertchian: One thing that could make this clearer is to pass the "should push" boolean as an input parameter. We did discuss it here: https://github.com/Alfresco/alfresco-build-tools/pull/71#discussion_r891976829 but this was postponed.
I can do the change now (as well as put all secrets as inputs, so that I don't update all the multiple callers of this reusable work multiple times).

gionn: I read the issue but not the comments, sorry for that 🤐 now it's clear!

gionn: > "Require branches to be up to date before merging" is not really helping

true, imagine if a bunch of dependabot PR get opened in the same batch, you are going to rebase/merge the remaining PRs after each automerge (and spend a lot of build minutes for that)

and unluckily dependabot doesn't support [updating multiple dependencies in a single PR](https://github.com/dependabot/dependabot-core/issues/1190#issuecomment-1071367790)

at this point I would suggest disabling that option if is generating more headaches than value and/or try to investigate some other tools (recently I am looking at [updatecli](https://www.updatecli.io/)

atchertchian: Thanks for suggesting so, that would certainly help.
Now to test the behavior I would need that PR to be merged first. Are you ok keeping that check as is (again, the same condition is used multiple times in this workflow already)?

gionn (approved):

PR #97, opened 2022-07-22
add custom message support and special case for issues

alxgomz:
Allow custome message in block when posting to slack (required for OPSEXP-1623)

### DIFF
--- a/.github/actions/send-slack-notification/action.yml
+++ b/.github/actions/send-slack-notification/action.yml
@@ -50,15 +50,16 @@ runs:
       run: |
         case $EVENT_NAME in
           pull_request)
-            echo "::set-output name=result::${BLOCK_MESSAGE:-${{ github.event.pull_request.title }}}"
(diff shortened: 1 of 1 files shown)


.github/actions/send-slack-notification/action.yml
```diff
@@ -43,13 +46,19 @@ runs:
...
+          pull_request)
+            echo "::set-output name=result::${BLOCK_MESSAGE:-${{ github.event.pull_request.title }}}"
+            ;;
+          issues)
+            echo "::set-output name=result::${BLOCK_MESSAGE:-${{ github.event.issue.body }}}"
+            ;;
+          *)
+            echo "::set-output name=result::${BLOCK_MESSAGE:-${{ github.event.head_commit.message }}}"
+            ;;
+        esac
```
gionn, line 61: would work as intended anyway, but I think it's more clear if before the switch we just check if BLOCK_MESSAGE is not empty and just use it

gionn (approved):

atchertchian (commented):
Nice! :+1:

PR #105, opened 2022-07-29
AAE-10135: turn build-and-tag-maven GH reusable workflow into an acti…

atchertchian:
…on

### DIFF
--- a/.github/actions/maven-build-and-tag/action.yml
+++ b/.github/actions/maven-build-and-tag/action.yml
@@ -15,10 +15,10 @@ inputs:
   extra-maven-opts:
     description: Extra maven options for the build command line
     required: false
-  nexus-username:
+  maven-username:
     description: Nexus user name
     required: true
-  nexus-password:
+  maven-password:
     description: Nexus password
     required: true
   quay-username:
@@ -62,8 +62,8 @@ runs:
         property-to-update: ${{ inputs.property-to-update }}
         maven-cli-opts: --show-version --no-transfer-progress --settings settings.xml
       env:
-        MAVEN_USERNAME: ${{ inputs.nexus-username }}
-        MAVEN_PASSWORD: ${{ inputs.nexus-password }}
+        MAVEN_USERNAME: ${{ inputs.maven-username }}
+        MAVEN_PASSWORD: ${{ inputs.maven-password }}
 
     - name: Set version env variable
       if: github.event_name == 'push'
@@ -106,8 +106,8 @@ runs:
       run: mvn ${{ steps.define_maven_command.outputs.command }} ${{ env.MAVEN_CLI_OPTS}} ${{ inputs.extra-maven-opts }}
       env:
         MAVEN_CLI_OPTS: --show-version --no-transfer-progress --settings settings.xml -Dlogging.root.level=off -Dspring.main.banner-mode=off -Ddocker.skip
-        MAVEN_USERNAME: ${{ inputs.nexus-username }}
-        MAVEN_PASSWORD: ${{ inputs.nexus-password }}
+        MAVEN_USERNAME: ${{ inputs.maven-username }}
+        MAVEN_PASSWORD: ${{ inputs.maven-password }}
 
     - name: Echo Longest Tests run
       shell: bash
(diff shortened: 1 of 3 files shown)


atchertchian:
Will rework the PR as some of the changes were already handled with #114

gionn (approved):

atchertchian:
force pushed to handle conflict after merge of #126

erdemedeiros (approved):
LGTM!

atchertchian:
force pushed to handle conflict after latest release

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -0,0 +1,148 @@
...
+    required: false
+  extra-maven-opts:
+    description: Extra maven options for the build command line
+    required: false
+  nexus-username:
+    description: Nexus user name
+    required: true
+  nexus-password:
+    description: Nexus password
+    required: true
```
gionn, line 23: maybe rename to maven since nexus it's the name of the product we are using for hosting maven repos

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -0,0 +1,148 @@
...
+  quay-password:
+    description: Quay.io password
+    required: true
+  docker-username:
+    description: Docker.io user name
+    required: true
+  docker-password:
+    description: Docker.io password
+    required: true
```
gionn, line 35: maybe those could be optional since it depends on the packages you are building if they need to push to quay or docker or both

you could also skip docker-login steps depending on those inputs empty or not (otherwise they will fail)

erdemedeiros: IIRC, we always need to login to Docker Hub, even when we are pushing images to quay.io, that's because some base images and images used by integration tests are hosted in Docker Hub and without login, we run out of quota to pull those images.

erdemedeiros (approved):
LGMT!

.github/actions/maven-build-and-tag/action.yml
atchertchian: That would make sense but right now there is more convention towards nexus than maven naming...:
```
$ git grep nexus-username 
.github/actions/maven-build-and-tag/action.yml:  nexus-username:
.github/actions/maven-build-and-tag/action.yml:        MAVEN_USERNAME: ${{ inputs.nexus-username }}
.github/actions/maven-build-and-tag/action.yml:        MAVEN_USERNAME: ${{ inputs.nexus-username }}
.github/actions/maven-release/action.yml:  nexus-username:
.github/actions/maven-release/action.yml:        NEXUS_USERNAME: "${{ inputs.nexus-username }}"
.github/actions/nexus-create-staging/action.yml:  nexus-username:
.github/actions/nexus-create-staging/action.yml:        NEXUS_USERNAME: ${{ inputs.nexus-username }}
.github/workflows/build-and-tag-maven.yml:          nexus-username: ${{ secrets.NEXUS_USERNAME }}
```
```
$ git grep maven-username 
.github/actions/maven-deploy-file/action.yml:  maven-username:
.github/actions/maven-deploy-file/action.yml:        MAVEN_USERNAME: ${{ inputs.maven-username }}
.github/actions/maven-deploy-file/action.yml:            echo "::error ::You need to provide credentials as maven-username and maven-password inputs when not providing a custom settings.xml"
.github/actions/maven-deploy-file/action.yml:        MAVEN_USERNAME: ${{ inputs.maven-username }}
```

atchertchian: made the change nevertheless

.github/actions/maven-build-and-tag/action.yml
gionn: yeah for your specific use case it's fine but I am always thinking if there is more room to make this action reusable in different contexts

erdemedeiros: It makes sense to make it as reusable as possible, but I prefer to minimize the amount of work without anticipating too much. The risk is spending time adding logic that will never be used. We can always evolve the implementation to cover new scenarios when they are needed. Does it make sense?

gionn (approved):

atchertchian:
Keeping the work branch for now as there are currently running workflows that are still running on it

PR #106, opened 2022-08-10
OPSEXP-1545 Decrease duplication in build scripts for helm charts with github custom actions

kklassa:
This PR introduces two new build actions (for now) - verify-compose which is to be used in [acs-deployment](https://github.com/Alfresco/acs-deployment) and verify helm for both acs-deployment and [alfresco-process-services-deployment](https://github.com/Alfresco/alfresco-process-services-deployment). Whereas the verify-compose seems to be working just fine (see run https://github.com/Alfresco/acs-deployment/actions/runs/2830976158), it is not as simple with verify-helm - it works in aps repo (see run https://github.com/Alfresco/alfresco-process-services-deployment/actions/runs/2826113606) but not in acs repo (example in acs run linked earlier), even though it is pretty much the same workflow, only the script differs and I copied the script straight from the respective repos. As a side note I thought about merging both the scripts as they contain a lot of duplicated code, but I decided that making such a merged script work for both repos would require a lot of conditional statements and that would hinder the readability of the script.


kklassa:
https://github.com/Alfresco/acs-deployment/actions/runs/2831908496 -> after env cleanup and removing `-e` flag from shebang all checks are passing and the cluster is clean ✅

.github/actions/verify-compose/action.yml
```diff
@@ -0,0 +1,54 @@
+name: "Verify and test docker compose"
```
gionn, line 1: if all the steps of this actions make a workflow, just declare a reusable workflow instead of an action https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow

.github/actions/verify-compose/docker_compose.sh
```diff
@@ -0,0 +1,105 @@
...
+  echo "BRANCH_NAME variable is not set"
+  exit 2
+fi
+
+GIT_DIFF=$(git diff origin/master --name-only .)
+compose_file="docker-compose.yml"
+alf_port=8080
+
+if [[ ${ACS_VERSION} != "latest" ]]; then
+  export compose_file="${ACS_VERSION}-docker-compose.yml"
```
gionn, line 21: maybe change the input of the action and expect to provide a full path to the docker compose file to test

.github/actions/verify-helm/aps_helm_install.sh
```diff
@@ -0,0 +1,239 @@
+#!/bin/bash
```
gionn, line 1: I think this file is just an outdated version of acs_helm_install, try to get rid of this one

kklassa:
I added two small actions - get-branch-name and get-commit-message because even though it's one step they were duplicated in many places - and they seem to work fine: https://github.com/Alfresco/alfresco-process-services-deployment/runs/7767082479?check_suite_focus=true

.github/actions/verify-helm/aps_helm_install.sh
kklassa: For now I use the helm_install.sh script form acs-deployment for both, but  I needed to add a check to skip some tests which were in the acs version but not the aps one, because they were failing. For now  both should work the same as before migrating the actions to this repo.

.github/actions/verify-compose/docker_compose.sh
kklassa: I did that, now the action verify-compose expects full path to the docker compose file. Run on which it worked: https://github.com/Alfresco/acs-deployment/actions/runs/2838100486

.github/actions/verify-compose/action.yml
kklassa: well in the original workflow in acs-deployment this makes up just one job out of four, and that workflow differs from the aps-deployment so I don't think making it a workflow will be beneficial for our current usecase.

kklassa:
In current configuration both the verify-helm and verify-compose run as expected - in acs-deployment: https://github.com/Alfresco/acs-deployment/actions/runs/2838322826 and in alfresco-process-services-deployment: https://github.com/Alfresco/alfresco-process-services-deployment/actions/runs/2838179178

kklassa:
For now I "dry" tested the publish-chart action, meaning that all the github pushing and tagging is commented out but I don't expect it to not work (hopefully), this is the run that shows the script gets executed and variables are propagated correctly: https://github.com/Alfresco/alfresco-process-services-deployment/runs/7784899710?check_suite_focus=true

kklassa:
These are all the changes I wanted to introduce, but before the eventual merge I still need to change the github ref in places where get-brach-name and get-commit-message are called within other actions (from OPSEXP-1545-helm-build-actions to main(?)) and  uncomment the actual script in publish_chart.sh.

.github/actions/get-branch-name/action.yml
```diff
@@ -0,0 +1,10 @@
+name: "Get branch name"
```
gionn, line 1: if you move this and get-commit-message in a separated PR we can merge it in advance so you can have a working reference (`@master`?)

kklassa: I created https://github.com/Alfresco/alfresco-build-tools/pull/107

gionn (approved):

.github/actions/publish-chart/action.yml
```diff
@@ -0,0 +1,32 @@
...
+  using: composite
+  steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Get branch name
+      uses: Alfresco/alfresco-build-tools/.github/actions/get-branch-name@OPSEXP-1545-helm-build-actions
+    - name: Get commit msg
+      uses: Alfresco/alfresco-build-tools/.github/actions/get-commit-message@OPSEXP-1545-helm-build-actions
```
gionn, line 24: switch them after merging https://github.com/Alfresco/alfresco-build-tools/pull/108

gionn (changes requested):

kklassa:
I think all works now finally - successful build on acs-deployment: https://github.com/Alfresco/acs-deployment/actions/runs/3015373352 and on alfresco-process-services-deployment: https://github.com/Alfresco/alfresco-process-services-deployment/actions/runs/3008039724

gionn (commented):
I think we should wrap all those actions under a common folder because (really no idea about the name):
* they address a specific need and there is little to no reusability beyond the 2 existing repos
* on the main actions folder there are some very similar actions coming from AAE (that should probably be moved under a folder as well)

@alxgomz your take on this?

alxgomz:
> I think we should wrap all those actions under a common folder because (really no idea about the name):
> 
>     * they address a specific need and there is little to no reusability beyond the 2 existing repos
> 
>     * on the main actions folder there are some very similar actions coming from AAE (that should probably be moved under a folder as well)
> 
> 
> @alxgomz your take on this?

Make sense indeed.  The name could reflect the fact these are just for acs/aps charts (like `dbp-charts`). Anyway I'm exploring using the already existing build-tools helm actions  as part of OPSEXP-1447 (which requires reviewing the build & release process)

kklassa:
I moved the three actions to `.github/actions/dbp-charts/` directory, I also considered `.github/dbp-charts/actions/` but I decided it makes more sense this way, please correct me if I am wrong 😅

kklassa:
green builds, at last 🙏
acs: https://github.com/Alfresco/acs-deployment/actions/runs/3022309397
aps: https://github.com/Alfresco/alfresco-process-services-deployment/actions/runs/3021408176

gionn (approved):

PR #109, opened 2022-08-23
ACS-3436 Add configure-git-author and veracode actions

damianujma:
This PR adds two new custom Github actions: 
- `configure-git-author` action can be used to configure the git username and email to associate commits with the provided identity
- `veracode` action can be used to run Veracode Source Clear Scan


.github/actions/config-git/action.yml
```diff
@@ -0,0 +1,18 @@
...
+  user-email:
+    description: "Email to link commits to user"
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - shell: bash
+      run: git checkout -B ${{ inputs.branch-name }}
+    - shell: bash
+      run: git fetch
```
gionn, line 16: checkout is usually done best via the standard actions/checkout

in this action I would just address the user.email and user.name configuration (so I would call it something like configure-git-author)

gionn (changes requested):
also add a brief section for both new actions in https://github.com/Alfresco/alfresco-build-tools#github-actions-provided-by-us

the veracode action can be proposed as a replacement of .travis.veracode.yml in the migration table (https://github.com/Alfresco/alfresco-build-tools#migrate-from-travis-to-github-actions)

.github/actions/config-git/action.yml
damianujma: Done.

gionn (approved):

PR #110, opened 2022-08-23
OPSEXP-1693 Add action to deploy arbitrary files to nexus

gionn:
OPSEXP-1693


kklassa (approved):

README.md
```diff
@@ -246,6 +262,30 @@ Commits local changes after configuring git user and showing the status of what
         commit-message: "My commit message"
 ```
 
+### maven-deploy-file
+
+Upload one or more files to a maven server, without requiring the presence of a
+pom.xml. It provides a settings.xml when can find one already configured. When
+using a custom settings.xml, you probably want to provide also
+```repository-id``` that match the credentials id to be used for deploying.
```
gionn, line 270:
```suggestion
`repository-id` that match the credentials id to be used for deploying.
```

alxgomz (approved):

PR #113, opened 2022-08-25
Acs 3436 fix source clear script

damianujma:
This is a PR to fix a bug in Veracode Source Clear Scan script and to add an information about alternatives to Travis CI default environment variables in `README.md`

### DIFF
--- a/.github/actions/veracode/source_clear.sh
+++ b/.github/actions/veracode/source_clear.sh
@@ -3,8 +3,6 @@
 echo "=========================== Starting SourceClear Script ==========================="
 PS4="\[\e[35m\]+ \[\e[m\]"
 set +e -v -x
-pwd
-pushd "$(dirname "${BASH_SOURCE[0]}")/../../"
 
 mvn -B -q clean install \
     -DskipTests \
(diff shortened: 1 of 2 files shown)


dsibilio (approved):

.github/actions/veracode/source_clear.sh
```diff
@@ -3,7 +3,6 @@
 echo "=========================== Starting SourceClear Script ==========================="
 PS4="\[\e[35m\]+ \[\e[m\]"
 set +e -v -x
-pushd "$(dirname "${BASH_SOURCE[0]}")/../"
```
dsibilio, line 6: Since we removed the `pushd` instruction here, shouldn't we also remove the `popd` at L#21, or what purpose does it serve at this point?

dsibilio (commented):
Since

gionn (approved):

.github/actions/veracode/source_clear.sh
damianujma: Good point, we don't need the `popd` instruction, I removed it.

PR #114, opened 2022-08-29
AAE-8254 add actions needed to migrate activiti-scripts

erdemedeiros:
Actions needed to migrate activiti-scripts and that can be reused by alfresco-process-scripts were moved here.

### DIFF
--- a/.github/actions/parse-next-release-from-chart/action.yml
+++ b/.github/actions/parse-next-release-from-chart/action.yml
@@ -1,0 +1,33 @@
+name: Parse next release from chart
+description: >
+  Parses the next main release version based on the content of Chart.yaml.
+  The suffix `-SNAPSHOT` is removed.
+inputs:
+  chart-dir:
+    description: "The relative path to the directory holding Chart.yaml"
+    required: true
+  chart-repository-dir:
+    description: "The path the git repository holding the chart project"
+    required: false
+outputs:
+  next-release:
+    description: Next released parsed from the Chart.yaml file
+    value: ${{ steps.parse-next-release.outputs.next-release }}
+runs:
+  using: composite
+  steps:
+    - name: Parse next release
+      id: parse-next-release
+      shell: bash
+      env:
+        CHART_REPO_DIR: ${{ inputs.chart-repository-dir }}
+      run: |
+        if [ -n "$CHART_REPO_DIR" ]
+        then
+          cd $CHART_REPO_DIR
+        fi
+
+        NEXT_VERSION=$(yq e '.version' ${{ inputs.chart-dir }}/Chart.yaml | grep -o "[0-9]*\.[0-9]*.[0-9]*")
+        echo "Next final version: $NEXT_VERSION"
+        echo "::set-output name=next-release::$NEXT_VERSION"
+
--- a/.github/actions/release-maven/action.yml
+++ b/.github/actions/release-maven/action.yml
@@ -1,0 +1,154 @@
+name: Release maven project
+description: Release Maven project
+
+inputs:
+  repo:
+    description: "Repository to be released"
+    required: true
+  base-tag:
+    description: "The name of the tag from where the release will be created"
+    required: true
+  extra-replacements:
+    description: Extra properties to replace
+    required: false
+  release-version:
+    description: "The name of the version to be released"
+    required: true
+  staging-repository-id:
+    description: "The id of the staging repository where the release artifacts will be stored"
+    required: true
+  github-token:
+    description: "The token to used to clone the repositories"
+    required: true
+  git-username:
+    description: "The username for git commit"
+    required: true
+  gpg-passphrase:
+    description: "GPG passphrase"
+    required: true
+  gpg-secret-keys:
+    description: "GPG secrets key"
+    required: true
+  gpg-owner-trust:
+    description: "GPG Owner trust"
+    required: true
+  nexus-username:
+    description: "Nexus user name"
+    required: true
+  nexus-password:
+    description: "Nexus password"
+    required: true
+  nexus-url:
+    description: "Base URL to nexus server"
+    required: false
+    default: "https://artifacts.alfresco.com/nexus"
+  maven-settings-path:
+    description: "path to the settings.xml to be used for the release"
+    required: false
+    default: ${{ github.workspace }}/maven-config/settings.xml
+
+runs:
+  using: composite
+  steps:
+    - name: set global properties
+      shell: bash
+      env:
+        REPO_DIR: ${{ github.workspace }}/${{ inputs.repo }}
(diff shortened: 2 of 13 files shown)


.github/actions/release-maven/action.yml
```diff
@@ -0,0 +1,154 @@
...
+    default: "https://artifacts.alfresco.com/nexus"
+  maven-settings-path:
+    description: "path to the settings.xml to be used for the release"
+    required: false
+    default: ${{ github.workspace }}/maven-config/settings.xml
+
+runs:
+  using: composite
+  steps:
+    - name: set global properties
```
atchertchian, line 53:
```suggestion
    - name: Set global properties
```

.github/actions/release-maven/action.yml
```diff
@@ -0,0 +1,154 @@
...
+    - name: Commit changes
+      if: steps.check-tag.outputs.exists == 'false'
+      uses: Alfresco/alfresco-build-tools/.github/actions/git-commit-changes@v1.12.0
+      with:
+        username: ${{ inputs.git-username }}
+        add-options: -u
+        commit-message: "Release version ${{ inputs.release-version}}"
+        repository-directory: ${{ env.REPO_DIR }}
+
+    - name: create tag
```
atchertchian, line 148:
```suggestion
    - name: Create and push tag
```

.github/actions/update-chart-version/action.yml
```diff
@@ -5,14 +5,23 @@ inputs:
     description: "The new version of the chart"
     required: true
   chart-dir:
-    description: "The path to the directory holding Chart.yaml"
+    description: "The relative path to the directory holding Chart.yaml"
     required: true
+  chart-repository-dir:
```
atchertchian, line 10: same question here: why can't we just have one input with the path to the directory holding this file?

.github/actions/release-maven/update-pom-version.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
```
atchertchian, line 2: I'm not sure I follow the need for such a script: seems to me that the sed replacement could replace unrelated version that happen to match the regexp, while using the maven versions:set command would handle this more smoothly.
What am I missing?

atchertchian (changes requested):
I did a few suggestions and added some comments.
I think I'm missing some context, maybe viewing usage in https://github.com/Activiti/activiti-scripts/pull/51 would help me understand some of the use cases (?)
Also please consider updating the README to document the new actions.

.github/actions/release-maven/action.yml
```diff
@@ -0,0 +1,154 @@
+name: Release maven project
```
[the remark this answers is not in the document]
erdemedeiros: It makes sense, I'll rename it.

.github/actions/release-maven/update-pom-version.sh
erdemedeiros: Before backporting the sed replacements herem, I tried to use the `mvn versions:set ` command, but it didn't work out of the box. I preferred to don't spend too much time on it as we can always improve that later on and the amount of work is already huge without it. 
One of the replacements that are not working out of the box with `mvn versions:set` is [this one](https://github.com/Activiti/activiti-cloud-application/blob/develop/activiti-cloud-dependencies/pom.xml#L8).

atchertchian: Thanks for the explanation (and I'm not saying this should be done in the context of that PR): it makes sense that this version would not be replaced given the fact that this project is in another repository at https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-build/pom.xml#L7. This version should be handled as a dependency via a property in the root pom.
This maven management should definitely be reviewed as I see this as a big risk for a release to fail (or not represent correctly what we're releasing): this sed replacement might not be reliable.

erdemedeiros: I agree, I don't like `sed` here neither. I don't think you can reference the parent version as a property, but it should be possible to review the module's inheritance.

.github/actions/parse-next-release-from-chart/action.yml
```diff
@@ -0,0 +1,32 @@
+name: Parse next release from chart
+description: >
+  Parses the next main release version based on the content of Chart.yaml.
+  The suffix `-SNAPSHOT` is removed.
+inputs:
+  chart-dir:
+    description: "The relative path to the directory holding Chart.yaml"
+    required: true
+  chart-repository-dir:
```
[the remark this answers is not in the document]
erdemedeiros: Good catch! Actually, I should be able to keep only `chart-repository-dir` and use `yq e '.version' **/Chart.yaml` in place of `yq e '.version' ${{ inputs.chart-dir }}/Chart.yaml`. I'll update it.

.github/actions/helm-parse-next-release/action.yml
```diff
@@ -0,0 +1,29 @@
...
+      shell: bash
+      env:
+        CHART_REPO_DIR: ${{ inputs.chart-repository-dir }}
+      run: |
+        if [ -n "$CHART_REPO_DIR" ]
+        then
+          cd $CHART_REPO_DIR
+        fi
+
+        NEXT_VERSION=$(yq e '.version' **/Chart.yaml | grep -o "[0-9]*\.[0-9]*.[0-9]*")
```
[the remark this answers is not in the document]
atchertchian: confusion --> resolving, sorry

.github/actions/parse-next-release-from-chart/action.yml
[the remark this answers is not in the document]
erdemedeiros: In the end, I think it's better to keep it like that. The alternatives would be either using `*/**/Chart.yaml` to match the file, but it less specific. Or add some logic to find one of the information based on the other one, but I'd prefer to avoid that.

.github/actions/update-chart-version/action.yml
erdemedeiros: Same as before.

README.md
```diff
@@ -342,6 +435,17 @@ Publishes a new helm chart package (`.tgz`) to a helm chart repository
           token: ${{ secrets.BOT_GITHUB_TOKEN}}
 ```
 
+### release-descriptor-loader
```
atchertchian, line 438: I should have commented on this earlier, sorry, but does it make sense to rename it to "load-release-descriptor"? Other actions due use a verb, and the "release" word can be misleading here.

atchertchian (approved):
Nice work! :+1:

README.md
erdemedeiros: It makes sense. Thanks. Fixed.

PR #123, opened 2022-09-21
OPSEXP-1545 Fix compose tests for forked PRs

kklassa:
Now the action relies on passing valid container registry credentials instead of checking the github event type.

### DIFF
--- a/.github/actions/dbp-charts/verify-compose/action.yml
+++ b/.github/actions/dbp-charts/verify-compose/action.yml
@@ -28,7 +28,7 @@ runs:
       run: |
         echo "${{ github.event.pull_request.head.repo.fork }}"
       shell: bash
-    - if: '! contains(inputs.compose_file_path, "community") && ! github.event.pull_request.head.repo.fork'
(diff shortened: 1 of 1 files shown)


.github/actions/dbp-charts/verify-compose/action.yml
```diff
@@ -24,21 +24,20 @@ runs:
...
       with:
         fetch-depth: 0
-    - if: ${{ inputs.acs_version != 'community' }} && ! github.event.pull_request.head.repo.fork
+    - if: ${{ inputs.quay_username != '' && inputs.quay_password != '' }}
       name: Login to Quay.io
       uses: docker/login-action@v1
       with:
         registry: quay.io
         username: ${{ inputs.QUAY_USERNAME }}
         password: ${{ inputs.QUAY_PASSWORD }}
```
gionn, line 33: check case on both login-action

gionn (changes requested):

gionn (approved):

PR #127, opened 2022-09-28
make docker hub creds and file optional

alxgomz:
Ref: OPSEXP-1746

### DIFF
--- a/.github/actions/dbp-charts/verify-compose/action.yml
+++ b/.github/actions/dbp-charts/verify-compose/action.yml
@@ -3,13 +3,14 @@ description: "Setup environment and run docker compose tests"
 inputs:
   compose_file_path:
     description: "full path to docker compose file to test"
-    required: true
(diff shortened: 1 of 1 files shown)


.github/actions/dbp-charts/verify-compose/action.yml
```diff
@@ -3,19 +3,22 @@ description: "Setup environment and run docker compose tests"
...
+    required: false
+    default: "./docker-compose.yml"
   quay_username:
     description: "username for Quay.io"
     required: false
   quay_password:
     description: "password for Quay.io"
     required: false
+  # docker hub credentials are only usefull to avoid rate limiting
+  # and must not be used in fork triggered workflow
```
gionn, line 15: that's true for our specific use case only - let's try to be generic here in alfresco-build-tools

kklassa (approved):

PR #128, opened 2022-09-28
OPSEXP-1746: allow passing additional args to pre-commit

alxgomz:
Ref: OPSEXP-1746

The goal is to use relevant pre-commit hooks in relevant workflows, but also offer any other option.
Also bumped the pre-commit to new major v3. Althoug hthis is a new major interface looks similar, they only moved from a js action to a composite one.

### DIFF
--- a/.github/actions/pre-commit/action.yml
+++ b/.github/actions/pre-commit/action.yml
@@ -28,7 +28,7 @@ runs:
         [ ${{ inputs.pre-commit-all-files }} ] && EXTRA_ARGS='--all-files'
         EXTRA_ARGS="${EXTRA_ARGS} ${{ inputs.pre-commit_args }}"
         echo "EXTRA_ARGS="${EXTRA_ARGS}" >> $GITHUB_ENV"
-    - uses: pre-commit/action@v2.0.3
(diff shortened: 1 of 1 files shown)


.github/actions/pre-commit/action.yml
```diff
@@ -11,7 +20,15 @@ runs:
...
       uses: actions/setup-python@v3
+    - name: build extra_args
+      shell: bash
+      env:
+        EXTRA_ARGS: ''
+      run: |
+        [ ${{ inputs.pre-commit-all-files }} ] && EXTRA_ARGS='--all-files'
+        EXTRA_ARGS="${EXTRA_ARGS} ${{ inputs.pre-commit_args }}"
+        echo "EXTRA_ARGS="${EXTRA_ARGS}" >> $GITHUB_ENV"
+    - uses: pre-commit/action@v2.0.3
```
kklassa, line 31: I see this action has a [v3.0.0](https://github.com/pre-commit/action/releases/tag/v3.0.0) version available, but I don't think it makes any difference for us here at all

kklassa (approved):

atchertchian (approved):

PR #136, opened 2022-10-07
OPSEXP-1746: helm skip co

alxgomz:
Ref: OPSEXP-1746

### DIFF
--- a/.github/actions/dbp-charts/verify-helm/action.yml
+++ b/.github/actions/dbp-charts/verify-helm/action.yml
@@ -50,12 +50,16 @@ inputs:
     description: "version of the kubectl CLI"
     required: false
     default: 'v1.23.6'
-
+  skip_checkout:
+    description: "Do not checkout the branch again"
+    required: false
 
 runs:
   using: composite
   steps:
     - name: Checkout
+      if: |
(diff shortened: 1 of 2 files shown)


gionn (approved):

.github/actions/dbp-charts/verify-helm/action.yml
```diff
@@ -50,12 +50,16 @@ inputs:
     description: "version of the kubectl CLI"
     required: false
     default: 'v1.23.6'
-
+  skip_checkout:
+    description: "Do not checkout the branch again"
+    required: false
```
gionn, line 55: add a default

gionn (changes requested):

.github/actions/dbp-charts/verify-helm/action.yml
alxgomz: I did not put one as the input var is only used in the `if` statement so result would be the same

gionn: I would never trusted github to implicitly default to false

gionn (approved):

gionn (approved):

PR #144, opened 2022-10-13
OPSEXP-1777 Fixup verify-helm action

gionn:
* have a different namespace when building a matrix, so parallelization is an option
* increase retry timeout for test - to give more time to flaky things to settle up and for manual inspection
* remove hard sleep before executing tests since we are already retrying them automatically
* remove unnecessary cleanup in pods_ready function
* bump ingress-nginx to version 4.2.5

OPSEXP-1777

### DIFF
--- a/.github/actions/dbp-charts/verify-helm/helm_install.sh
+++ b/.github/actions/dbp-charts/verify-helm/helm_install.sh
@@ -11,12 +11,10 @@ clean_up () {
 }
 trap clean_up EXIT
 
-GIT_DIFF="$(git diff origin/master --name-only .)"
-namespace=$(echo "${BRANCH_NAME}" | cut -c1-28 | tr /_ - | tr -d '[:punct:]' | awk '{print tolower($0)}')"-${GITHUB_RUN_NUMBER}"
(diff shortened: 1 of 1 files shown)


.github/actions/dbp-charts/verify-helm/helm_install.sh
```diff
@@ -11,13 +11,11 @@ clean_up () {
 }
```
alxgomz, line 11: we are deleting the ingress release but not the acs one. IS there a reason for that?

alxgomz (commented):
just one point

.github/actions/dbp-charts/verify-helm/helm_install.sh
gionn: I got fooled by that too, double check the helm delete argument(s)

do you think for clarity it's better to have two different helm delete commands, one for each release?

alxgomz: Ah ok, no it's OK now we know it :)

alxgomz (approved):

PR #146, opened 2022-10-14
fix: add retry support for jx-updatebot-pr action

igdianov:
This PR adds support for retrying `jx-updatebot-pr` action in case there is a race condition on the target pr branch due to a concurrent commit from another project, i.e.

```
- uses: Alfresco/alfresco-build-tools/.github/actions/jx-updatebot-pr@ref
  with:
    retries: '3'
    retries-wait: '10'

```

### DIFF
--- a/.github/actions/jx-updatebot-pr/action.yml
+++ b/.github/actions/jx-updatebot-pr/action.yml
@@ -81,11 +81,11 @@ inputs:
   retries:
     description: 'How many times to retry on failure'
     required: false
-    default: 3
+    default: '5'
   retries-wait:
     description: 'How many seconds to wait between retries'
     required: false
-    default: 5
+    default: '10'
 runs:
   using: composite
   steps:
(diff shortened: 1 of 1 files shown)


.github/actions/jx-updatebot-pr/action.yml
```diff
@@ -78,6 +78,14 @@ inputs:
     required: false
     description: project root directory. Defaults to '.'
     default: '.'
+  retries:
+    description: 'How many times to retry on failure'
```
atchertchian, line 82: With the term "retry", I feel like as a user, I would need to put "0" to avoid retrying (eg run only once), but given the implementation, it would need to be 1, correct?
If so, maybe the implementation should be changed to add + 1 to this input.
Also, I'm in favor of not retrying by default.

atchertchian: (but if the default is something we need to manage globally anyway, I'm ok keeping 5, it really depends on the reasons behind the need to retry here: can you elaborate on the problem please?)

gionn (commented):
in some projects we already started using [nick-fields/retry](https://github.com/nick-fields/retry#examples) community action for flaky run commands, I would prefer using that instead

gionn:
Closing due to inactivity

PR #162, opened 2022-10-25
OPSEXP-1806 Fix multiline/markdown support in slack notification

gionn:
add tests

OPSEXP-1806

### DIFF
--- a/.github/actions/send-slack-notification/compute-message.sh
+++ b/.github/actions/send-slack-notification/compute-message.sh
@@ -1,0 +1,22 @@
+#!/bin/bash -e
+
+echo 'result<<EOF'
+
+if [ -n "$BLOCK_MESSAGE" ]; then
+    echo -n "${BLOCK_MESSAGE}" | sed -z 's/\n/\\n/g'
+else
+    case $EVENT_NAME in
+        pull_request)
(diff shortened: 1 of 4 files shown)


gionn:
Screenshot 2022-10-25 at 12 11 17

atchertchian (approved):
:+1:

.github/actions/send-slack-notification/compute-message.sh
```diff
@@ -0,0 +1,22 @@
...
+        issues)
+        echo -n "${ISSUE_BODY}" | sed -z 's/\n/\\n/g'
+        ;;
+        *)
+        echo -n "${COMMIT_MESSAGE}" | sed -z 's/\n/\\n/g'
+        ;;
+    esac
+fi
+
+echo ''
```
alxgomz, line 21: what about removing the `-n` switch from `echo` commands  and removing that empty `echo` ?

alxgomz (approved):

.github/actions/send-slack-notification/compute-message.sh
gionn: it is to avoid having that last newline converted to `\n` by sed, while I really want a NL before the closing heredoc tag

`echo "\nEOF"` would have worked as well but for some reason looked more cleaner :D

PR #168, opened 2022-11-09
AAE-10136: fix helm publish workflow by downloading package previousl…

atchertchian:
…y produced

The helm-package-chart action is no longer putting the package in current directory -> it needs to be downloaded so that it can be used by the helm-publish-chart action.

### DIFF
--- a/.github/actions/helm-publish-chart/action.yml
+++ b/.github/actions/helm-publish-chart/action.yml
@@ -104,7 +104,7 @@ runs:
         then
           echo "result=$INPUT_MESSAGE" >> $GITHUB_OUTPUT
         else
-          RELEASE_PACKAGE_NAME=basename $RELEASE_PACKAGE_PATH
+          RELEASE_PACKAGE_NAME=$(basename $RELEASE_PACKAGE_PATH)
           echo "result=Release $RELEASE_PACKAGE_NAME" >> $GITHUB_OUTPUT
         fi
 
--- a/.github/actions/helm-publish-chart/action.yml
+++ b/.github/actions/helm-publish-chart/action.yml
@@ -4,6 +4,9 @@ inputs:
   chart-package:
     description: "The name of the packaged chart been added to the repository. I.e. common-x.y.z.tgz"
     required: true
+  chart-package-path:
+    description: "The path of the packaged chart been added to the repository. I.e. common-x.y.z.tgz. Used the package name if not set."
+    required: false
   helm-charts-repo:
     description: "The name of the repository where the package will be added"
     required: true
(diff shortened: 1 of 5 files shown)


.github/workflows/helm-publish-new-package-version.yml
```diff
@@ -76,6 +76,10 @@ jobs:
       - name: Push tag
         run: git push origin $VERSION
 
+      - uses: actions/download-artifact@v3
+        with:
+          name: ${{ steps.package-helm-chart.outputs.package-file }}
+
```
gionn, line 82: can't you expose the path as an output of helm-package-chart and use that instead of downloading again? you already have filename as an output

I know it's working as it is but `download-artifact` is really meant to be used to fetch an artifact produced in a previous job of a workflow - and ok that charts are small but you are consuming time and bandwidth for a thing that is already somewhere in the current workspace

atchertchian: I agree it would be a better fix but I'm confused by the multiple usage of this action, and the README documents that downloading the artifact is the proper way to retrieve it.
I'd need to review the multiple callers in all repos, so I was tempted to revert the changes instead (eg the package is produced in current folder, as previously assumed). I'm not sure why this assumption was not present on https://github.com/Alfresco/alfresco-build-tools/blob/master/.github/actions/helm-release-and-publish/action.yml#L79

atchertchian: @gionn I changed the logic, I will have to issue 4 more PRs to align on this (after release)

.github/actions/helm-publish-chart/action.yml
```diff
@@ -86,10 +89,16 @@ runs:
...
       run: |
           mkdir $RANDOM_PATH
-          RELEASE_PACKAGE_NAME=${{ inputs.chart-package }}
-          cp "$RELEASE_PACKAGE_NAME" $RANDOM_PATH
+          if [ -n "$RELEASE_PACKAGE_PATH" ]
+          then
+            cp "$RELEASE_PACKAGE_PATH" $RANDOM_PATH
+          else
+            cp "$RELEASE_PACKAGE_NAME" $RANDOM_PATH
+          fi
```
gionn, line 101: don't know if is going to break too many things, but you can drop `chart-package` input and just have the `chart-package-path` input and when you need the name of the file only you can extract it with `basename ${{ inputs.chart-package-path }}`

gionn: that could work too because in the end `file.ext` is still a valid path, but make sure to make it clear in the doc of the input that is accepting a path as well

.github/actions/helm-publish-chart/action.yml
```diff
@@ -86,25 +86,25 @@ runs:
...
       env:
         INPUT_MESSAGE: ${{ inputs.git-message }}
-        RELEASE_PACKAGE_NAME: ${{ inputs.chart-package }}
+        RELEASE_PACKAGE_PATH: ${{ inputs.chart-package }}
       run: |
         if [ -n "$INPUT_MESSAGE" ]
         then
           echo "result=$INPUT_MESSAGE" >> $GITHUB_OUTPUT
         else
+          RELEASE_PACKAGE_NAME=basename $RELEASE_PACKAGE_PATH
```
gionn, line 107: shouldn't be:
```suggestion
          RELEASE_PACKAGE_NAME=$(basename $RELEASE_PACKAGE_PATH)
```
?

gionn (approved):

PR #185, opened 2022-11-21
AAE-11403 experiment prettier format

atchertchian:


atchertchian:
Will add more to docs/readme once we agree on the PR content.

.github/actions/pre-commit-default/action.yml
```diff
@@ -0,0 +1,50 @@
...
+description: "Install and run pre-commit"
+inputs:
+  pre-commit-args:
+    description: Additionnal parameters to pass to pre-commit
+    required: false
+  pre-commit-all-files:
+    description: >-
+      Whether to run pre-commit checks on all files (defaults to "true")
+    required: false
+    default: "true"
```
gionn, line 11: I don't know if we really have an use case for when we don't want --all-files when running on CI

you got this by forking the existing pre-commit action right? maybe drop support for it until we figure out if it's really needed

.github/actions/pre-commit-default/action.yml
```diff
@@ -0,0 +1,50 @@
...
+  check-github-configuration:
+    description: >-
+      Whether to run github actions, workflows, dependabot pre-commit checks.
+    required: false
+    default: "true"
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/checkout@v3
```
gionn, line 30: not sure if we really want to checkout every time, this will impact workflows that are doing a previous actions/checkout with customized parameters (calling multiple times actions/checkout will just reset the current working tree)

gionn (commented):
I like the overall approach 💯

.github/actions/pre-commit-default/action.yml
atchertchian: I thought about it and then kept it because it still allows me to handle formatting only on changed files from the caller using something like:

```
      - uses: Alfresco/alfresco-build-tools/.github/actions/pre-commit-default@AAE-11403-experiment-prettier-format
        with:
          pre-commit-all-files: false
          pre-commit-args: --files ${{ steps.file_changes.outputs.files }}
```

Now I'm ok removing all this as diff support would be included inside the action (if needed later), wdyt?

.github/actions/pre-commit-default/action.yml
atchertchian: Would adding an input to avoid checkout (`do-checkout` true by default (?)) be enough?

erdemedeiros: That's a good point @gionn , and IMO, this is one of the hardest part of creating GH actions: find the right granularity to avoid code duplication without being too opinionated or needing to declare all the possible customisations that called actions can provide.

In this specific case, can we assume that pre-commits will always going to have their dedicated job? If that's the case, I don't think we'll need customisations on checkout, as the customisations will be probably be done in another job.

docs/README.md
```diff
@@ -656,6 +657,45 @@ or into an existing workflow of your choice just declaring the step:
...
+        with:
+          check-format: 'false'
+```
+
+Or:
+
+```yml
+      - uses: Alfresco/alfresco-build-tools/.github/actions/pre-commit-default@ref
+        with:
+          check-format: 'false'
```
erdemedeiros, line 696: Is this suppose to be `check-github-configuration`? Currently it's the same as the previous example.

erdemedeiros (changes requested):
The is small detail in the Readme. LGTM, otherwise.

.github/actions/pre-commit-default/action.yml
atchertchian: (pushed: removed the input for all files)

.github/actions/pre-commit-default/action.yml
atchertchian: It's true that what I'm about to do is in caller workflows (whether they're in a dedicated job or not) is:
- call pre-commit action (leveraging the local pre-commit-config file if any)
- call pre-commit-default action (so here the checkout would be unneeded)

I did condition this checkout to a `do-checkout` parameter (true by default)

docs/README.md
atchertchian: Thanks! fixed

erdemedeiros (dismissed):
LGTM!

.github/actions/pre-commit-default/action.yml
gionn: > this is one of the hardest part of creating GH actions

it is indeed and I started to prefer to be less DRY and avoiding doing too much things in a single custom action. In the end is not really a problem to add an additional `- uses: actions/checkout@v3` in every workflow before using your custom action.

I am not against `do-checkout` input but I am not sure that including actions/checkout in a custom action is really worth the value (I mean, you are just saving a line in yaml).

gionn (approved):
last comment here https://github.com/Alfresco/alfresco-build-tools/pull/185/files#r1035769921

anyway is fine for me

.github/actions/pre-commit-default/action.yml
atchertchian: I agree and after thinking about it, it would make sense to remove it completely. Now we'll have an inconsistency between `pre-commit` and `pre-commit-default` actions: should this be a concern?

.github/actions/pre-commit-default/action.yml
```diff
@@ -0,0 +1,50 @@
+name: "Pre-commit format"
+description: "Install and run pre-commit"
+inputs:
+  pre-commit-args:
+    description: Additionnal parameters to pass to pre-commit
+    required: false
+  pre-commit-all-files:
```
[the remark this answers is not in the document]
atchertchian: I renamed to `pre-commit-args` to `extra-args` after all because the inputs started diverging anyway, and because it's consistent with `extra-maven-opts` that we defined for the `maven-build-and-tag` action.

.github/actions/pre-commit-default/action.yml
atchertchian: After discussion, I removed the checkout and adjusted the README to make this explicit.

gionn (approved):

PR #186, opened 2022-11-22
fix(AAE-11371): update helm-release-and-publish input params 

igdianov:
This PR adds missing input params to helm-release-and-publish action to be able to reuse it in https://github.com/Alfresco/alfresco-process-infrastructure-deployment/pull/825

### DIFF
--- a/.github/actions/helm-release-and-publish/action.yml
+++ b/.github/actions/helm-release-and-publish/action.yml
@@ -29,6 +29,10 @@ inputs:
   git-username:
     description: 'The username to commit on the git repositories'
     required: true
+  do-push:
+    description: "If 'true', changes will be pushed ('false' can be used for PRs)"
+    default: 'true'
+    required: false
 
 runs:
   using: composite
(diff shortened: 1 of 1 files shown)


erdemedeiros (approved):
LGTM!

.github/actions/helm-release-and-publish/action.yml
```diff
@@ -17,6 +17,12 @@ inputs:
   helm-repository-branch:
     description: 'Branch on the charts repository'
     required: false
+  helm-repository-subfolder:
+    description: 'The name of the subfolder inside the charts repository where the package should be added'
+    required: false
+  helm-repository-base-url:
+    description: 'Helm chart repo base url'
+    required: true
```
atchertchian, line 25: I think it should not be required (it is not required in `helm-publish-chart` action either)

atchertchian (changes requested):
I think there's a mistake on the input requirement, please double-check.

Also I pushed a commit to be able to run this action in PRs (eg do everything but pushing what was done remotely), please check if that looks good

.github/actions/helm-release-and-publish/action.yml
igdianov: Fixed

igdianov:
> I think there's a mistake on the input requirement, please double-check.
> 
> Also I pushed a commit to be able to run this action in PRs (eg do everything but pushing what was done remotely), please check if that looks good

Looks good!

atchertchian (approved):

PR #187, opened 2022-11-22
AAE-11466: allow showing maven downloads on build-and-tag action

atchertchian:
The new `verbose` input should allow us to debug issues related to maven downloads.

### DIFF
--- a/.github/actions/maven-build-and-tag/action.yml
+++ b/.github/actions/maven-build-and-tag/action.yml
@@ -36,6 +36,10 @@ inputs:
   git-username:
     description: The username to commit on the git repository
     required: true
+  verbose:
+    description: Whether additional logs should be displayed (maven download logs for instance)
+    required: false
+    default: 'false'
 
 outputs:
   version:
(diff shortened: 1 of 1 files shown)


atchertchian:
Will rework this PR so that we still have a way to avoid showing downloads by default, but can easily show them

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -60,13 +64,26 @@ runs:
...
         distribution: ${{ inputs.java-distribution }}
 
+    - name: Compute maven options
+      id: compute-maven-options
+      shell: bash
+      env:
+        INPUT_VERBOSE: ${{ inputs.verbose }}
+      run: |
+        NTP=''
+        if [[ '$INPUT_VERBOSE' == 'true' ]]
```
gionn, line 74: double quotes or won't expand $INPUT_VERBOSE

atchertchian: Fixed, thanks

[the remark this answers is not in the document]
atchertchian: This is still buggy: the evaluation is `true` without quotes...

gionn (approved):

PR #194, opened 2022-12-06
AAE-11373 add possibility to skip docker build

erdemedeiros:
Acceptance tests does not produce any docker images, so we need to be able to skip this step


andrea-ligios (approved):
LGTM!

PR #198, opened 2022-12-09
AAE-11711: add text on slack notif payload

atchertchian:
The main text was missing from the payload: the API was generating warnings because of that:

```
[WARN]  web-api:WebClient:0 The top-level `text` argument is missing in the request payload for a chat.postMessage call - It's a best practice to always provide a `text` argument when posting a message. The `text` is used in places where the content cannot be rendered such as: system push notifications, assistive technology such as screen readers, etc.
```

I just moved existing content around to make the message more compact and closer to other github skack notifications.

Here's a sample test notification:

Screenshot from 2022-12-09 10-45-01

### DIFF
--- a/.github/actions/send-slack-notification/action.yml
+++ b/.github/actions/send-slack-notification/action.yml
@@ -87,6 +87,7 @@ runs:
         channel-id: "${{ inputs.channel-id }}"
         payload: |
           {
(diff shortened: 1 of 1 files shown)


erdemedeiros (approved):
LGTM! Thank you for the improvements!

PR #200, opened 2022-12-12
AAE-11722 ease up reportportal integration

atchertchian:
This PR adds two new actions to help with reportportal integration.
It also allows appending a custom message to the default one when sending a slack notification (useful in reportportal usage situations where we want to add information to link to reportportal launches).

There's heavy bash usage (and I'm thankful for bats) but I'm not fluent in bash: all feedback on bash commands is very welcome.

I will link several dependent PRs referencing this PR soon.


.github/actions/reportportal-prepare/tests/reportportal-prepare.bats
```diff
@@ -0,0 +1,148 @@
...
+    export RP_URL="https://rpserver:8080 with spaces too"
+    export RP_PROJECT="my project with spaces"
+
+    run get-rp-input.sh
+    [ "$status" -eq 0 ]
+
+    expected=$(cat << BATS
+enabled=true
+key=my launch prefix with spaces-weird event-3665876492
+url=https://rpserver:8080 with spaces too/ui/#my project with spaces/launches/all
```
mauriziovitale, line 139: I think it would be good to have the reposerver as env variable

mauriziovitale (approved):

.github/actions/reportportal-prepare/tests/reportportal-prepare.bats
atchertchian: It is the default value for these helper actions, but you can define as an env variable from the caller if it differs

PR #201, opened 2022-12-13
AAE-11760 git-latest-tag-sha action

ozeta:
this action finds the latest tag for a repository, and returns its long sha


.github/actions/git-latest-tag-sha/action.yml
```diff
@@ -0,0 +1,24 @@
...
+      run: |
+        if [ -n "$REPO_DIR" ]
+        then
+          cd $REPO_DIR
+        fi
+        git fetch --tags
+        last_tag=$(git tag --sort=-creatordate | head -n 1)
+        tag_long_sha=$(git rev-list -n 1 $last_tag)
+        echo "Long SHA for tag $last_tag is $tag_long_sha"
+        echo "tag_long_sha=$tag_long_sha" >> $GITHUB_OUTPUT
```
gionn, line 24: no need to duplicate the existing git-latest-tag, just extend that by adding a new output

gionn (changes requested):

.github/actions/git-latest-tag/tests/git-latest-tag.bats
```diff
@@ -0,0 +1,35 @@
...
+    export GITHUB_HEAD_REF="AAE-11760-get-latest-tag-sha"
+
+    # Mock git-latest-tag defaults
+    export PATTERN="*"
+    export REPO_DIR="$PWD"
+}
+
+@test "latest_tag" {
+    latest_tag=$(git tag --sort=-creatordate | head -n 1)
+    tag_sha=$(git rev-list -n 1 $latest_tag)
```
gionn, line 19: it doesn't make sense to use in the test the same code used in the code under test to check the behavior.

you could verify that the message is structured correctly and do not check any specific tag/sha (that will change every time a new release is made)

gionn (changes requested):

.github/actions/git-latest-tag/tests/git-latest-tag.bats
ozeta: yeah i tought that was an unuseful test and actually I was wait for a suggestion 😂

.github/actions/git-latest-tag/tests/git-latest-tag.bats
```diff
@@ -0,0 +1,36 @@
...
+    # Mock git-latest-tag defaults
+    export PATTERN="*"
+    export REPO_DIR="$PWD"
+}
+
+@test "latest_tag" {
+    regex="v[0-9]+\.[0-9]+\.[0-9]+"
+    run git-latest-tag.sh
+    [ "$status" -eq 0 ]
+    if [[ ! $output =~ $regex ]]; then
```
gionn, line 21: I would have checked the entire string via regex but there is also the other test so in the end is fine for me

gionn (approved):

PR #211, opened 2023-01-04
AAE-7305 add action to auto promote Nexus staging repository

erdemedeiros:


.github/actions/nexus-promote-staging/action.yml
```diff
@@ -0,0 +1,83 @@
...
+        CLOSE_STAGING_PAYLOAD_CONTENT=$(envsubst < $GITHUB_ACTION_PATH/close-payload-template.json)
+        curl -u "${NEXUS_USERNAME}":"${NEXUS_PASSWORD}" \
+          -H "Accept: application/json" \
+          -H "Content-Type: application/json" \
+          -d "$CLOSE_STAGING_PAYLOAD_CONTENT" \
+          "${NEXUS_URL}"/service/local/staging/bulk/close
+
+        wait_until_true () {
+            local attempt_counter=0
+            local max_attempts=12
```
atchertchian, line 41: Would it make sense to make this value a non-required input?

.github/actions/nexus-promote-staging/action.yml
```diff
@@ -0,0 +1,83 @@
...
+            do
+              if [ ${attempt_counter} -eq ${max_attempts} ]
+              then
+                echo "Max attempts reached"
+                exit 1
+              fi
+
+              attempt_counter=$((attempt_counter+1))
+              echo "Condition not reached yet. Attempt $attempt_counter out of $max_attempts. Retrying..."
+              sleep 5
```
atchertchian, line 53: Same, would it make sense to turn this hardcoded value into a non-required input?

.github/actions/nexus-promote-staging/action.yml
```diff
@@ -0,0 +1,83 @@
...
+            | yq -p=xml e '.stagingRepositories.data.stagingProfileRepository[]| select (.repositoryId == env(STAGING_REPOSITORY_ID)) | .type ' \
+            | grep open > /dev/null 2>&1
+        }
+
+        # the close operation is asynchronous and before proceeding with the release step, we need to make sure the repository is closed
+        echo "Waiting for staging repository $STAGING_REPOSITORY_ID to get closed..."
+        wait_until_true staging_repo_is_closed
+        echo "Repository $STAGING_REPOSITORY_ID successfully closed!"
+
+    - name: Release staging repository
```
atchertchian, line 69: The action name does not make it explicit that a release is also performed. We do not have any use case to promote without releasing, but maybe it would make sense to move release to a dedicated action, wdyt?
Also, it would allow skipping the async check if not needed (as the release operaton is async too: maybe the completion should be checked on it too?)

.github/actions/nexus-promote-staging/action.yml
```diff
@@ -0,0 +1,83 @@
...
+              echo "Condition not reached yet. Attempt $attempt_counter out of $max_attempts. Retrying..."
+              sleep 5
+            done
+        }
+
+        staging_repo_is_closed () {
+          curl -u "${NEXUS_USERNAME}":"${NEXUS_PASSWORD}" \
+            -X GET "${NEXUS_URL}"/service/local/staging/profile_repositories \
+            | yq -p=xml e '.stagingRepositories.data.stagingProfileRepository[]| select (.repositoryId == env(STAGING_REPOSITORY_ID)) | .type ' \
+            | grep open > /dev/null 2>&1
```
atchertchian, line 61: There is no direct API to get the status of the staging repo?

.github/actions/nexus-promote-staging/action.yml
erdemedeiros: yes, it makes sense. I'll add it and make the current value as default.

.github/actions/nexus-promote-staging/action.yml
atchertchian: I'm fine with it, I was just wondering.
If it's of any help, I found this in old release notes at https://help.sonatype.com/repomanager2/release-notes/older-release-notes:

```
close/promote/release/drop operations are now asynchronous. Drop operations, including nexus-staging:release with auto-drop enabled, are complete when the response code is 404. For other operations, a GET request to "/service/local/staging/repository/<repo-id>" that returns a 'transitioning" field value of false indicates the operation is complete.
```

.github/actions/nexus-promote-staging/action.yml
erdemedeiros: In all the use cases we have now, we always need to close and release, but I can split it into too actions and add option to wait for the close one:
```yaml
     - uses: Alfresco/alfresco-build-tools/.github/actions/nexus-close-staging@ref
        with:
          wait-for-completion: true
          version: ${{ inputs.version }}
          staging-repository: ${{ inputs.staging-repository}}
          nexus-username: ${{ secrets.NEXUS_USERNAME }}
          nexus-password: ${{ secrets.NEXUS_PASSWORD }}
    
     - uses: Alfresco/alfresco-build-tools/.github/actions/nexus-release-staging@ref
        with:
          version: ${{ inputs.version }}
          staging-repository: ${{ inputs.staging-repository}}
          nexus-username: ${{ secrets.NEXUS_USERNAME }}
          nexus-password: ${{ secrets.NEXUS_PASSWORD }}
```

Concerning the wait mechanism, I only added for the first one because we need to make sure the staging repository is closed before calling the release action. BTW, the naming is not consistent between the UI and the related REST API, as in the UI the `Release` button calls the `promote` API.
For the second one, the wait mechanism is not strictly needed, but it could be added for consistency and if we want to make sure that the repository was actually been released. The check however, would mean verifying that the repository has disappeared from the list of repositories.

What do you think?

.github/actions/nexus-promote-staging/action.yml
```diff
@@ -0,0 +1,83 @@
...
+      shell: bash
+      env:
+        NEXUS_USERNAME: ${{ inputs.nexus-username }}
+        NEXUS_PASSWORD: ${{ inputs.nexus-password }}
+        NEXUS_URL: ${{ inputs.nexus-url }}
+        VERSION: ${{ inputs.version }}
+        STAGING_REPOSITORY_ID: ${{ inputs.staging-repository }}
+      run: |
+        PROMOTE_PAYLOAD_CONTENT=$(envsubst < $GITHUB_ACTION_PATH/promote-payload-template.json)
+        curl -u "${NEXUS_USERNAME}":"${NEXUS_PASSWORD}" \
```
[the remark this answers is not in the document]
erdemedeiros: Thank you for the suggestion, I'll add it.

.github/actions/nexus-promote-staging/action.yml
erdemedeiros: Ok, I'll update it. Thanks for the feedback.

.github/actions/nexus-promote-staging/action.yml
erdemedeiros: indeed, this helps: there is an API to retrieve directly the repository using `/service/local/staging/repository/{repositoryId}`. However, in the case of the release I still prefer to use the one returning a list of repositories to avoid handling `404` in the case where the release is completed.

atchertchian (approved):

PR #215, opened 2023-01-09
[AAE-11861] add 'Create preview' feature

jsokolowskii:
When a PR is created with `preview` label, the build creates and publishes maven versions and docker images for the project with a snapshot tag `0.0.1-${GITHUB_PR_NUMBER}-${GITHUB_RUN_NUMBER}-SNAPSHOT`. Tested e.g. on https://github.com/Alfresco/alfresco-deployment-service/pull/996

### DIFF
--- a/.github/actions/maven-build-and-tag/action.yml
+++ b/.github/actions/maven-build-and-tag/action.yml
@@ -74,7 +74,7 @@ runs:
 
     - name: Set preview version
       id: set-preview-version
-      if: github.event_name == 'push' && ${{ contains(github.head_ref, 'preview') }}
+      if: github.event_name == 'push' && ${{ contains(github.event.pull_request.labels.*.name, 'Create preview') }}
       shell: bash
       run: |
         GITHUB_PR_NUMBER=PR-${{ github.event.pull_request.number }}
@@ -127,7 +127,8 @@ runs:
           echo "command=verify" >> $GITHUB_OUTPUT
         fi
       env:
-        DO_PUSH: ${{ github.event_name == 'push' }}
+        DO_PUSH: ${{ github.event_name == 'push' || ( github.event_name == 'pull_request' && \
+          contains(github.event.pull_request.labels.*.name, 'Create preview') ) }}
 
     - name: Build and Test with Maven (and maybe Deploy)
       shell: bash
@@ -150,7 +151,8 @@ runs:
           echo "command=--push" >> $GITHUB_OUTPUT
         fi
       env:
-        DO_PUSH: ${{ github.event_name == 'push' }}
+        DO_PUSH: ${{ github.event_name == 'push' || ( github.event_name == 'pull_request' && \
+          contains(github.event.pull_request.labels.*.name, 'Create preview') ) }}
 
     - name: Docker Build (and maybe Push)
       shell: bash
--- a/.github/actions/maven-build-and-tag/action.yml
+++ b/.github/actions/maven-build-and-tag/action.yml
@@ -72,9 +72,17 @@ runs:
         ${{ inputs.verbose }} && NTP=''
         echo "result=--show-version --settings settings.xml $NTP" >> $GITHUB_OUTPUT
 
+    - name: Set is_preview env variable
+      id: set-is-preview-env-variable
+      env:
+        IS_PREVIEW: ${{ github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'Create preview') }}
+      shell: bash
+      run: |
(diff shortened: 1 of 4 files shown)


.github/actions/maven-build-and-tag/action.yml
```diff
@@ -72,27 +72,44 @@ runs:
         ${{ inputs.verbose }} && NTP=''
         echo "result=--show-version --settings settings.xml $NTP" >> $GITHUB_OUTPUT
 
+    - name: Set is_preview env variable
+      id: set-is-preview-env-variable
+      env:
+        IS_PREVIEW: ${{ github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'Create preview') }}
```
atchertchian, line 78: I wonder if "preview" could not be enough as a label (as it's already a verb), and actually, maybe this should be extracted as an input so that it can integrate better with the caller repository pull request labels, wdyt?

atchertchian (commented):
I'm ok with the changes, it's a nice improvement! Please update the README file to advertize this feature.

I do have concerns about the cleanup on the maven and docker artifacts: maybe some refinement could be helpful here to help with cleanup (at least on docker images, adding dedicated labels, for instance).

Also, see my suggestion about making it possible to override the PR label in case caller repositories want to use something other than "Create preview" (or just "preview" if you're ok with it)

.github/actions/maven-build-and-tag/action.yml
jsokolowskii: @atchertchian I added the input and changed the default label to `preview`. Thanks for the comment!

jsokolowskii:
> I'm ok with the changes, it's a nice improvement! Please update the README file to advertize this feature.
> 
> I do have concerns about the cleanup on the maven and docker artifacts: maybe some refinement could be helpful here to help with cleanup (at least on docker images, adding dedicated labels, for instance).
> 
> Also, see my suggestion about making it possible to override the PR label in case caller repositories want to use something other than "Create preview" (or just "preview" if you're ok with it)

@atchertchian thank you. I added a section in README.
About the cleanup, I think it could be done by filtering out the tags with the pattern `0.0.1-$GITHUB_PR_NUMBER-$GITHUB_RUN_NUMBER-SNAPSHOT`. Adding a second tag to a docker image would mean changing all the `build-and-push-docker-images.sh` files in the target repositories to allow an additional tag (or multiple tags). The maven artifacts could be cleaned up the same way I suppose.

docs/README.md
```diff
@@ -598,6 +598,12 @@ Check out, builds a maven project and docker images, generating a new alpha vers
           git-username: ${{ secrets.BOT_GITHUB_USERNAME }}
 ```
 
+#### Preview option for maven-build-and-tag
+
+There is a possibility to publish snapshot maven artifacts and docker images from an open PR. 
+In order to use it specify `preview-label` input (or use default `preview`). Create a PR with the `preview-label` label.
+The created artifacts will be tagged as `0.0.1-$GITHUB_PR_NUMBER-$GITHUB_RUN_NUMBER-SNAPSHOT`.
```
atchertchian, line 605: I wonder if the run number should be used here: do we want to keep dedicated snapshots per run? If so, maybe using the run id instead would be better (?)
And I'm not sure this versioning pattern applies to deployed docker image tags: can you please document those too?

jsokolowskii: I think we may settle on PR number only to avoid multiple snapshots.
As for the docker images, the patter is the same, I'll add it to README.

jsokolowskii: Done.

atchertchian (approved):
LGTM!

.github/actions/maven-update-pom-version/action.yml
```diff
@@ -16,7 +16,7 @@ inputs:
 outputs:
   next-prerelease:
     description: "Next prerelease"
-    value: ${{ steps.next-prerelease-resolver.outputs.next-prerelease }}
+    value: ${{ steps.resolve-version.outputs.version }}
```
erdemedeiros, line 19: I think this is a left over from a previous change. There is no step with id `resolve-version` in this action, so all this block can be removed.

erdemedeiros (changes requested):
Just one small detail about a left over output. LGTM otherwise.

.github/actions/maven-update-pom-version/action.yml
jsokolowskii: Removed the whole block, thank you for spotting this!

erdemedeiros (approved):
LGTM!

PR #223, opened 2023-01-24
ACS-3750 Set default veracode-fails-build to 'true'

dsibilio:
I think it'd be safer to err on the side of caution and cause the builds to fail by default if the Veracode scan reports critical vulnerabilities or fails completely.

Also, given that the default value is to ignore failures, people trying to adopt the Veracode action often don't even realize that they're missing critical components to make the Veracode action work _(e.g.: the `SRCCLR_API_TOKEN` secret may be missing in the repository configuration, but it won't be very evident as the job goes green anyways)_.

It would be great to release this change before we start introducing the Veracode action into more repositories, so it won't be as impacting.

### DIFF
--- a/.github/actions/veracode/action.yml
+++ b/.github/actions/veracode/action.yml
@@ -6,7 +6,7 @@ inputs:
     required: true
   veracode-fails-build:
     description: "Determine if the veracode scan result should fail the build"
-    default: "false"
+    default: "true"
     required: false
 runs:
   using: "composite"
--- a/.github/actions/veracode/action.yml
+++ b/.github/actions/veracode/action.yml
@@ -4,10 +4,6 @@ inputs:
   srcclr-api-token:
     description: "Agent API Token"
     required: true
-  veracode-fails-build:
-    description: "Determine if the veracode scan result should fail the build"
-    default: "true"
-    required: false
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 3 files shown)


damianujma (dismissed):

gionn (commented):
If you feel brave enough I think we could completely get rid of this feature and suggest the usage of [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) that should also have better ux (step would be marked as red instead of green but the build will continue and overall status would be green)

dsibilio:
> If you feel brave enough I think we could completely get rid of this feature and suggest the usage of [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) that should also have better ux (step would be marked as red instead of green but the build will continue and overall status would be green)

It sounds like a good idea, I can't see any issues with it at the moment. I'll refactor the script and the action to completely remove the `veracode-fails-build` parameter and suggest the `continue-on-error` approach instead within the example.

damianujma (approved):

dsibilio:
Example of run **without** `continue-on-error: true`: https://github.com/Alfresco/alfresco-enterprise-repo/actions/runs/3997431631/jobs/6858776411
Example of run **with** `continue-on-error: true`: https://github.com/Alfresco/alfresco-enterprise-repo/actions/runs/3997534617/jobs/6859011011

Seems to be working as expected 👍

gionn (approved):

PR #226, opened 2023-01-30
ACS-4455 Install custom settings.xml as part of setup-java-build

dsibilio:
Installs a custom `settings.xml` file in the Maven home directory. Does not fail in case the file is not found, as not all repositories may need to install the custom file.

**Sample run**: https://github.com/Alfresco/alfresco-enterprise-repo/pull/1221

### DIFF
--- a/.github/actions/setup-java-build/action.yml
+++ b/.github/actions/setup-java-build/action.yml
@@ -25,7 +25,9 @@ runs:
           ${{ runner.os }}-maven-
     - name: "Install ${{ inputs.maven-settings }}"
       shell: bash
-      run: "cp ${{ inputs.maven-settings }} $HOME/.m2/settings.xml || :"
+      run: >
+        mkdir $HOME/.m2
+        cp ${{ inputs.maven-settings }} $HOME/.m2/settings.xml || true
     - name: "Set up Java"
       uses: actions/setup-java@v3
       with:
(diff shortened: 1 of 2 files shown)


damianujma (dismissed):

.github/actions/setup-java-build/action.yml
```diff
@@ -19,8 +23,12 @@ runs:
         key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
         restore-keys: |
           ${{ runner.os }}-maven-
+    - name: "Install ${{ inputs.maven-settings }}"
+      shell: bash
+      run: "cp ${{ inputs.maven-settings }} $HOME/.m2/settings.xml || :"
```
gionn, line 28: add a `mkdir $HOME/.m2/` otherwise will fail when there is not a cache to be restored

I don't see `:` used since long time, prefer the built-in `true` for more clarity

gionn (changes requested):

.github/actions/setup-java-build/action.yml
dsibilio: Done, thanks!

damianujma (approved):

aonych (approved):

gionn (approved):

PR #234, opened 2023-02-07
ACS-4500 Fix skip tests in reusable workflow

dsibilio:
Testing skip tests functionality in reusable workflow revealed that it doesn't work as intended. Possibly related to https://github.com/actions/runner/issues/1483.


dsibilio:
This is not needed, the issue was due to the consumer of the reusable workflow rather than the workflow itself.

PR #239, opened 2023-02-22
OPSEXP-2004 Add auto-commit support for pre-commit action

gionn:
Ref: OPSEXP-2004

Tested in https://github.com/Alfresco/alfresco-helm-charts/pull/8

### DIFF
--- a/.github/actions/pre-commit/action.yml
+++ b/.github/actions/pre-commit/action.yml
@@ -9,16 +9,16 @@ inputs:
       Whether to run pre-commit checks on all files (defaults to "true")
     required: false
     default: "true"
-  python-version:
-    description: The python version of the local runner
-    required: false
-    default: "3.9"
   auto-commit:
     description: >-
       Enables auto-commit of eventual fixups
       (requires `permissions.contents: write` on the calling job)
     required: false
     default: "false"
+  python-version:
+    description: The python version of the local runner
+    required: false
+    default: "3.9"
 runs:
   using: "composite"
   steps:
@@ -54,7 +54,7 @@ runs:
       id: auto-commit-action
       if: inputs.auto-commit == 'true'
 
-    - name: Re-throw potential pre-commit failure with no changes detected for auto-commit
+    - name: Re-throw potential pre-commit failure when no changes detected for auto-commit
       if: steps.pre-commit.outcome == 'failure' && (inputs.auto-commit != 'true' || steps.auto-commit-action.outputs.changes_detected == 'false')
-      run: echo "pre-commit failed with no changes detected for auto-commit" && exit 1
+      run: echo "pre-commit failed and no changes detected for auto-commit" && exit 1
       shell: bash
--- a/.github/actions/pre-commit/action.yml
+++ b/.github/actions/pre-commit/action.yml
@@ -2,7 +2,7 @@ name: "Pre-commit run"
 description: "Install and run pre-commit"
 inputs:
   pre-commit-args:
-    description: Additionnal parameters to pass to pre-commit
+    description: Additional parameters to pass to pre-commit
     required: false
   pre-commit-all-files:
     description: >-
@@ -52,7 +52,7 @@ runs:
       id: auto-commit-action
       if: inputs.auto-commit == 'true'
 
-    - name: rethrow failure when autocommit is disable or no changes has been detected
+    - name: Re-throw potential pre-commit failure with no changes detected for auto-commit
       if: steps.pre-commit.outcome == 'failure' && (inputs.auto-commit != 'true' || steps.auto-commit-action.outputs.changes_detected == 'false')
-      run: echo "pre-commit failed, auto-commit is disabled or no changes detected for auto-commit" && exit 1
+      run: echo "pre-commit failed with no changes detected for auto-commit" && exit 1
       shell: bash
--- a/.github/actions/pre-commit/action.yml
+++ b/.github/actions/pre-commit/action.yml
@@ -13,6 +13,10 @@ inputs:
     description: The python version of the local runner
     required: false
     default: "3.9"
+  auto-commit:
+    description: "enables auto-commit of eventual fixups (requires `permissions.contents: write` on the calling job)"
+    required: false
+    default: "false"
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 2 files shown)


.github/actions/pre-commit/action.yml
```diff
@@ -13,14 +13,22 @@ inputs:
...
+  auto-commit:
+    description: "enables auto-commit of eventual fixups (requires `permissions.contents: write` on the calling job)"
+    required: false
+    default: "false"
 runs:
   using: "composite"
   steps:
     - uses: actions/checkout@v3
+      with:
+        ref: ${{ github.head_ref }}
```
atchertchian, line 25: I'm not 100% sure about this: I understand this is needed for the auto-commit behavior, but I wonder if that changes what pre-commit is checking on PRs, because it would match the source branch (maybe I'm confused, just want to make sure of this)

.github/actions/pre-commit/action.yml
```diff
@@ -29,6 +37,18 @@ runs:
         ${{ inputs.pre-commit-all-files }} && EXTRA_ARGS='--all-files'
         EXTRA_ARGS="${EXTRA_ARGS} ${{ inputs.pre-commit-args }}"
         echo EXTRA_ARGS="${EXTRA_ARGS}" >> $GITHUB_ENV
+
     - uses: pre-commit/action@v3.0.0
+      id: pre-commit
+      continue-on-error: true
```
atchertchian, line 43:
```suggestion
      continue-on-error: ${{ inputs.auto-commit == 'true' }}
```

Not sure if that actually works, but I'd rather avoid re-throwing if auto-commit is disabled anyway, wdyt?

.github/actions/pre-commit/action.yml
```diff
@@ -29,6 +37,18 @@ runs:
...
       with:
         extra_args: ${{ env.EXTRA_ARGS }}
+
+    - uses: stefanzweifel/git-auto-commit-action@v4
+      id: auto-commit-action
+      if: inputs.auto-commit == 'true'
+
+    - name: rethrow failure when autocommit is disable or no changes has been detected
```
atchertchian, line 51:
```suggestion
    - name: Re-throw potential pre-commit failure when auto-commit is disabled or no changes have been detected
```
(to be adapted if continue-on-error suggestion is accepted)

.github/actions/pre-commit/action.yml
```diff
@@ -29,6 +37,18 @@ runs:
...
+      if: steps.pre-commit.outcome == 'failure' && (inputs.auto-commit != 'true' || steps.auto-commit-action.outputs.changes_detected == 'false')
+      run: echo "pre-commit failed, auto-commit is disabled or no changes detected for auto-commit" && exit 1
```
atchertchian, line 53: (if continue-on-error suggestion is valid and done)

```suggestion
      run: echo "pre-commit failed and no changes detected for auto-commit" && exit 1
```

.github/actions/pre-commit/action.yml
```diff
@@ -13,14 +13,22 @@ inputs:
     description: The python version of the local runner
     required: false
     default: "3.9"
+  auto-commit:
+    description: "enables auto-commit of eventual fixups (requires `permissions.contents: write` on the calling job)"
```
atchertchian, line 17:
```suggestion
    description: >-
      Enables auto-commit of eventual fixups
      (requires `permissions.contents: write` on the calling job)
```

and nitpicking: this input would be better placed before the "python-version" input

Also, since I do have the behavior on some workflows :sweat_smile: afaik only the first fixups are done and all hooks after are not run, so the pre-commit command would need to be executed multiple times to get all the potential fixups... (I considered adding an integer to control how many times it should be run and gave up: called this action multiple times continuing on error instead)

.github/actions/pre-commit/action.yml
gionn: should only affects workflow triggered by pull_request events (in all the other cases this var is empty - thus like not specifying any ref).

The difference is that pull_request by default run on the merge commit with the base branch (e.g. to see what happen once merged to main) and that is a detached head so you will lost any commit done on top of that.

adding head_ref make sure that git checkout the source branch and we get the expected behavior (additional commits appearing in the PR branch).

.github/actions/pre-commit/action.yml
gionn: looks cool indeed, let me try it.

.github/actions/pre-commit/action.yml
atchertchian: I gathered that this would not be an issue for push events, I was just not sure we'd run pre-commit with additional commits on the PR branch, thanks for confirming :+1:

.github/actions/pre-commit/action.yml
atchertchian: I was actually looking for such an option and did not find the "fail-fast" one, thanks!
Now if it's disabled by default, it's probably not going to fix my issue, but I'll look into it :+1:

atchertchian: Actually the fail fast option was activated... perfect!

.github/actions/pre-commit/action.yml
```diff
@@ -17,10 +23,16 @@ runs:
   using: "composite"
   steps:
     - uses: actions/checkout@v3
+      env:
+        REF_TO_CHECKOUT: ${{ inputs.auto-commit == 'true' && github.head_ref || '' }}
+      with:
+        ref: ${{ env.REF_TO_CHECKOUT }}
+
```
gionn, line 30: it's ugly to use an env but debugging is more easy

in this way we continue to checkout the merged head for PR when not using auto-commit

.github/actions/pre-commit/action.yml
```diff
@@ -29,6 +41,20 @@ runs:
         ${{ inputs.pre-commit-all-files }} && EXTRA_ARGS='--all-files'
         EXTRA_ARGS="${EXTRA_ARGS} ${{ inputs.pre-commit-args }}"
         echo EXTRA_ARGS="${EXTRA_ARGS}" >> $GITHUB_ENV
-    - uses: pre-commit/action@v3.0.0
+
+    - uses: pre-commit/action@646c83fcd040023954eafda54b4db0192ce70507 # v3.0.0
+      id: pre-commit
+      env:
+        CONTINUE_ON_ERROR: ${{ inputs.auto-commit == 'true' }}
+      continue-on-error: ${{ fromJSON(env.CONTINUE_ON_ERROR) }}
```
gionn, line 49: uglier than everything but it's the only way I found to make it work

.github/actions/pre-commit/action.yml
```diff
@@ -17,10 +23,16 @@ runs:
...
   steps:
     - uses: actions/checkout@v3
+      env:
+        REF_TO_CHECKOUT: ${{ inputs.auto-commit == 'true' && github.head_ref || '' }}
+      with:
+        ref: ${{ env.REF_TO_CHECKOUT }}
+
     - name: Set up Python ${{ inputs.python-version }}
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v4
```
atchertchian, line 32: This will allow us to get rid of the warnings, right?

atchertchian (approved):
Nice improvement!

PR #244, opened 2023-03-06
OPSEXP-1853: add an helm plugin action

alxgomz:
Ref: OPSEXP-1853

The helm3-unittest plugin we introduced is now superseeded by the helm-unitest one. Instead of introducing a feature flip to choose which unittest plugin I prefer introducing a generic plugin install action and deprecate the old one.

### DIFF
--- a/.github/actions/helm-plugin/action.yml
+++ b/.github/actions/helm-plugin/action.yml
@@ -1,0 +1,22 @@
+---
+name: Helm Plugin install
+description: >
+  Install an helm plugin from provided url
+inputs:
+  plugin_url:
+    description: URL where to get the plugin from
+    required: true
+  plugin_version:
+    description: version of the helm-unittest plugin to install
+    required: false
+    default: latest
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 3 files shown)


.github/actions/helm-plugin/action.yml
```diff
@@ -0,0 +1,22 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Install Helm plugin ${{ inputs.plugin_url }}
+      shell: bash
+      run: |
+        set -e
+        [ ${{ inputs.plugin_version }} != 'latest' ] && \
+        VER="--version ${{ inputs.plugin_version }}"
+        helm plugin install ${{ inputs.plugin_url }} $VER
```
gionn, line 22: not sure it's really worth having a custom action just for this but no strong opinion on that

gionn (approved):

.github/actions/helm-plugin/action.yml
alxgomz: yeah, now that I read it again... 
Though it allows deprecating the old action.

PR #249, opened 2023-03-27
HXOR-196 - Get pr number 2

maurizio-cacace:


.github/actions/get-pr-number/index.js
gionn, line 1: is this an exact copy of https://github.com/kamatama41/get-pr-number-action/blob/master/index.js?

gionn (changes requested):

maurizio-cacace:
PR not necessary anymore

PR #251, opened 2023-03-28
fix commit message doc

maurizio-cacace:


docs/README.md
```diff
@@ -387,6 +387,14 @@ Loads the content of the last commit message that triggered the action into `COM
       - uses: Alfresco/alfresco-build-tools/.github/actions/get-commit-message@ref
 ```
 
+To use the action in a proper way, be sure to use a checkout with fetch-depth option as follow:
```
gionn, line 390:
```suggestion
This action requires a checkout with fetch-depth option as follow:
```

I would merge the two snippet in a single one (beware of missing dash in the checkout step)

gionn (changes requested):

PR #255, opened 2023-04-26
AFI-1520:Externalize required parameters for build in build-and-relea…

SushmitaGoswami:
We need to externalize the following parameters for building the sap-connector projects,
1. java-version
2. java-distribution
3. release-version
4. next-dev-version.

While incorporating the required change, I found an issue of GitHub action while differentiating the input parameter between empty or not set in a reusable workflow and it is already an open issue at GitHub. Here is the reference -  [https://github.com/actions/runner/issues/924](https://github.com/actions/runner/issues/924 )

Hence, I needed to do it using condition construct. 

Thank you!

### DIFF
--- a/.github/workflows/build-and-release-maven.yml
+++ b/.github/workflows/build-and-release-maven.yml
@@ -7,10 +7,12 @@ on:
         description: the desired Java version
         type: string
         required: false
+        default: "17"
       java-distribution:
         description: the desired Java distribution
         type: string
         required: false
+        default: "temurin"
       skip-tests:
         description: skips the tests during the Build phase if set to 'true'
         type: boolean
@@ -36,12 +38,8 @@ on:
         type: string
         required: false
         default: ''
-      release-version:
-        description: the name of the version to be released
-        type: string
-        required: false
-      next-dev-version:
-        description: the name of the version to be set as next development version
+      extra-maven-opts:
+        description: additional Maven build options
         type: string
         required: false
 
@@ -67,15 +65,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: "Setup Java Build with specific configuration"
-        if: inputs.java-version != '' && inputs.java-distribution != ''
-        uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v1.36.0
+      - uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v1.36.0
         with:
           java-version: ${{ inputs.java-version}}
           java-distribution: ${{ inputs.java-distribution}}
-      - name: "Setup Java Build with default configuration"
-        if: inputs.java-version == '' || inputs.java-distribution == ''
-        uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v1.36.0
       - name: "Build"
         if: ${{ inputs.skip-tests }}
         run: mvn -B -V install -DskipTests ${{ inputs.build-args }}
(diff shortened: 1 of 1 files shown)


SushmitaGoswami:
Here is the link of the CI file of our project - [https://github.com/Alfresco/sap-content-connector-parent/blob/feature/AFI-568/.github/workflows/ci.yml](https://github.com/Alfresco/sap-content-connector-parent/blob/feature/AFI-568/.github/workflows/ci.yml)

This file is just for some testing at this moment. And also, for testing purpose, we created a local copy of the build-and-release-maven.yml which is referenced from the ci.yml at this moment. 

Thanks in advance!

.github/workflows/build-and-release-maven.yml
```diff
@@ -91,12 +114,24 @@ jobs:
...
+        uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v1.36.0
       - uses: Alfresco/alfresco-build-tools/.github/actions/configure-git-author@v1.36.0
         with:
           username: ${{ secrets.BOT_GITHUB_USERNAME }}
           email: ${{ secrets.BOT_GITHUB_EMAIL }}
       - name: "Build"
         run: mvn -B -V install -DskipTests ${{ inputs.build-args }}
-      - name: "Release"
+      - name: "Release with specific version"
+        if: inputs.release-version && inputs.next-dev-version
```
dsibilio, line 133: Why is this syntax different than the usual (missing the `!= ''` comparisons that are otherwise present in the other if conditions)?

dsibilio (changes requested):
Since this action is being used in several repositories, it would be very important to have a link to an example workflow run where we don't pass any of the newly added parameters to verify that it still works as expected _(installs the default Java version, computes the release and development version automatically)_ and this change won't break pre-existing workflows that get bumped to the latest version of this action.

.github/workflows/build-and-release-maven.yml
```diff
@@ -91,12 +114,24 @@ jobs:
...
+        run: mvn -B -DscmCommentPrefix="[maven-release-plugin][skip ci] " -Dusername="${{ secrets.BOT_GITHUB_USERNAME }}" -Dpassword="${{ secrets.BOT_GITHUB_TOKEN }}" -DskipTests -Darguments="-DskipTests ${{ inputs.release-args }}" -DreleaseVersion="${{ inputs.release-version }}" -DdevelopmentVersion="${{ inputs.next-dev-version }}" release:clean release:prepare release:perform
```
gionn, line 134: you can avoid duplicating the entire command here, if you add a step before that will accumulate the additional `-Dprop` you need in a new variable, that will default to empty string and which you can use straight in the existing release step (let me know if you need additional pointers for achieving this)

gionn (changes requested):

.github/workflows/build-and-release-maven.yml
```diff
@@ -91,12 +114,24 @@ jobs:
...
-      - uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v1.36.0
+      - name: "Setup Java Build with specific configuration"
+        if: inputs.java-version && inputs.java-distribution
+        uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v1.36.0
+        with:
+          java-version: ${{ inputs.java-version}}
+          java-distribution: ${{ inputs.java-distribution}}
+      - name: "Setup Java Build with default configuration"
+        if: inputs.java-version == '' || inputs.java-distribution == ''
+        uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v1.36.0
```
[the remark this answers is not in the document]
SushmitaGoswami: Actually, I didn't initialize the `java-version` and `java-distribution` to the default values since I thought if we update the default values of these properties in the `setup-java-build` action, then we would have to update it in this file as well. But if  that is okay, I will update these to their default values in this file as well. Please share your thoughts.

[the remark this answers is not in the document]
alxgomz: I also prefer to have a defaults set here so we can switch when we want to not when the upstream decides to roll it out.

.github/workflows/build-and-release-maven.yml
SushmitaGoswami: Sure, I will add a link to an example workflow which uses this workflow with the default values of those newly added params. Although, I have already tested it by changing my existing workflow and it is working fine. But I will add one for visibility and confirmation.

And regarding - _Why is this syntax different than the usual (missing the `!= ''` comparisons that are otherwise present in the other if conditions)?_
Actually, I checked that this was working in case the input parameters are non empty or set when I didn't mention `!= ''` for comparison explicitly. But if mentioning this explicitly improves the readability of the code, I am absolutely okay with this. Please share your thoughts, I will change it accordingly. 

Thank you.

.github/workflows/build-and-release-maven.yml
SushmitaGoswami: Sure. I will update it.

.github/workflows/build-and-release-maven.yml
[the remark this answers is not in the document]
SushmitaGoswami: Okay sure. I will update to the default values.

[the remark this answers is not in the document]
atchertchian: Agreed, I had the same thought :)

.github/workflows/build-and-release-maven.yml
atchertchian: Agreed with @gionn about avoid duplication, and also, I would go into a different direction and just offer additional maven props as an input, so that you can be more free to add custom additional parameters to the maven command line.

Said differently, instead of adding explicit inputs `release-version` and `next-dev-version` to the workflow, a generic `extra-maven-opts` option could make sense here, like it's done on https://github.com/Alfresco/alfresco-build-tools/blob/v1.36.0/.github/actions/maven-build-and-tag/action.yml#L144-L148 (with input defined at https://github.com/Alfresco/alfresco-build-tools/blob/v1.36.0/.github/actions/maven-build-and-tag/action.yml#L15-L17

SushmitaGoswami: @atchertchian I completely agree with your points and will change it accordingly.

SushmitaGoswami:
Updated the PR as per the review comments. 

Following are the sample workflows which consume this build-and-release-maven workflow.

1. **Link to CI file (with custom value of the properties)** - https://github.com/Alfresco/sap-content-connector-parent/blob/feature/AFI-1516/.github/workflows/release.yml
2. **Link to the CI file (with default value of the new properties)** - 
>- **CI file** - https://github.com/Alfresco/sap-content-connector-parent/blob/feature/AFI-1516/.github/workflows/release-test-only.yml
>- **sample build** - https://github.com/Alfresco/sap-content-connector-parent/actions/runs/4872878348

Please let me know if any additional change is required.

alxgomz (dismissed):
LGTM

.github/workflows/build-and-release-maven.yml
```diff
@@ -51,6 +66,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v1.36.0
+        with:
+          java-version: ${{ inputs.java-version}}
+          java-distribution: ${{ inputs.java-distribution}}
```
gionn, line 71:
```suggestion
          java-version: ${{ inputs.java-version }}
          java-distribution: ${{ inputs.java-distribution }}
```

gionn (commented):
just some missing spaces, then LGTM

dsibilio (dismissed):

.github/workflows/build-and-release-maven.yml
SushmitaGoswami: Done

gionn (approved):

dsibilio (approved):

PR #259, opened 2023-05-22
Feature/afi 1530

SushmitaGoswami:
We need to incorporate the installation process of toolchains.xml during the build process of a few of the alfresco SAP connector projects.

I have added two following parameters,
1. **configure-toolchains** (Added in order to keep the backward compatibility for those projects which don't need the toolchains.xml file)
2. **maven-toolchains**

Repository which is using alfresco-build-tools actions are as follows. I have tested the flow and it is working fine.
1. [https://github.com/Alfresco/sap-connector-build-tools/blob/feature/AFI-1530/.github/actions/build-and-test-maven/action.yml](https://github.com/Alfresco/sap-connector-build-tools/blob/feature/AFI-1530/.github/actions/build-and-test-maven/action.yml)
2. [https://github.com/Alfresco/sap-connector-build-tools/blob/feature/AFI-1530/.github/workflows/release-and-dryrun.yml](https://github.com/Alfresco/sap-connector-build-tools/blob/feature/AFI-1530/.github/workflows/release-and-dryrun.yml)

Sample project which need the toolchains.xml is as follows,
[https://github.com/Alfresco/sap-content-connector-accelerator/tree/feature/AFI-1530](https://github.com/Alfresco/sap-content-connector-accelerator/tree/feature/AFI-1530)

### DIFF
--- a/.github/actions/setup-java-build/toolchains.xml
+++ b/.github/actions/setup-java-build/toolchains.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<toolchains
+	xmlns="http://maven.apache.org/TOOLCHAINS/1.1.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/TOOLCHAINS/1.1.0 http://maven.apache.org/xsd/toolchains-1.1.0.xsd">
+	<toolchain>
+		<type>jdk</type>
+		<provides>
+			<version>17</version>
+		</provides>
+		<configuration>
+			<jdkHome>${env.JAVA_HOME}</jdkHome>
+		</configuration>
+	</toolchain>
+</toolchains>
--- a/.github/actions/setup-java-build/toolchains.xml
+++ b/.github/actions/setup-java-build/toolchains.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<toolchains
+<toolchains>
 	xmlns="http://maven.apache.org/TOOLCHAINS/1.1.0"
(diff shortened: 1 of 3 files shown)


.github/actions/setup-java-build/toolchains.xml
```diff
@@ -0,0 +1,14 @@
...
+	<toolchain>
+		<type>jdk</type>
+		<provides>
+			<version>17</version>
+		</provides>
+		<configuration>
+			<jdkHome>${env.JAVA_HOME}</jdkHome>
+		</configuration>
+	</toolchain>
+</toolchains>
```
gionn, line 14: AFAIK maven should already autodetect the default JDK by using JAVA_HOME and we never had to configure a toolchains.xml file also when using different java versions in our existing builds

which is the actual issue you were facing that lead to this change? can you provide a branch with a build failing due to the lack of this file?

gionn (changes requested):
which is the actual issue that you

.github/actions/setup-java-build/toolchains.xml
gionn: Thanks for being super clear!

I understand that you had this issue because the parent project you are using (maintained by a third party) uses the [maven-toolchain-plugin](https://maven.apache.org/plugins/maven-toolchains-plugin/) which requires a toolchains.xml setup to not fail.

IMHO the `maven-toolchains-plugin` is not so commonly used in these days and I literally cannot find any usage in the entire alfresco codebase, hence I would prefer to not increase the complexity of our `setup-java-build` action to cover this not really common use case.

I suggest that you provide it directly inside the workflows of the builds that requires it, and only if you need to handle it in the same way in different repositories, then you can define your own action and reuse it within your different repositories (ping me in chat if you need more guidance on this)

SushmitaGoswami:
As discussed closing this PR as the change is no longer needed. Thank you so much @gionn once again for your help.

PR #266, opened 2023-05-29
ACS-5201: ARM64 ACS Packaging CI failure

krdabrowski:
- Added new action docker-dump-containers-logs allowing to store Docker containers logs.

### DIFF
--- a/.github/actions/docker-dump-containers-logs/action.yml
+++ b/.github/actions/docker-dump-containers-logs/action.yml
@@ -1,0 +1,21 @@
+name: "Docker dump containers logs"
+description: "Persists Docker container logs"
+runs:
+  using: composite
+  steps:
+    - name: "Dump containers logs and archive them"
+      shell: bash
+      run: |
+        # Get containers state
+        docker ps -a
+        # Dump all containers logs to files
+        docker ps -a --format '{{.ID}} {{.Names}}' | while read -r id name ; do
+          docker logs "${id}" --tail 5000 > "${name}.log"
+        done
+        # Create archive with log files
+        tar -zcf logs.tar.gz *.log
+    - name: "Upload archive containing all *.log files"
+      uses: actions/upload-artifact@v3
+      with:
+        path: logs.tar.gz
+        name: logs.tar.gz
--- a/.github/actions/docker-dump-containers-logs/action.yml
+++ b/.github/actions/docker-dump-containers-logs/action.yml
@@ -20,7 +20,7 @@ runs:
         # Create archive with log files
         tar -zcf logs.tar.gz *.log
         # Specify output artefact name
-        if [ "${{ inputs.output-archive-name }}" -eq "containers-logs.tar.gz" ]; then
+        if [ "${{ inputs.output-archive-name }}" = "containers-logs.tar.gz" ]; then
           echo "artefactName=containers-logs-${{ github.job }}-${{ github.run_attempt }}.tar.gz" >> $GITHUB_ENV
         else
           echo "artefactName=${{ inputs.output-archive-name }}" >> $GITHUB_ENV
--- a/.github/actions/docker-dump-containers-logs/action.yml
+++ b/.github/actions/docker-dump-containers-logs/action.yml
@@ -21,7 +21,7 @@ runs:
         tar -zcf logs.tar.gz *.log
         # Specify output artefact name
         if [ "${{ inputs.output-archive-name }}" = "containers-logs.tar.gz" ]; then
-          echo "artefactName=containers-logs-${{ github.job }}-${{ github.run_attempt }}.tar.gz" >> $GITHUB_ENV
+          echo "artefactName=containers-logs-${{ github.job }}-${{ github.run_attempt }}-$(date +%Y%m%d%H%M%S).tar.gz" >> $GITHUB_ENV
         else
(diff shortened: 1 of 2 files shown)


dsibilio (dismissed):

.github/actions/docker-dump-containers-logs/action.yml
```diff
@@ -0,0 +1,21 @@
...
+        docker ps -a --format '{{.ID}} {{.Names}}' | while read -r id name ; do
+          docker logs "${id}" --tail 5000 > "${name}.log"
+        done
+        # Create archive with log files
+        tar -zcf logs.tar.gz *.log
+    - name: "Upload archive containing all *.log files"
+      uses: actions/upload-artifact@v3
+      with:
+        path: logs.tar.gz
+        name: logs.tar.gz
```
gionn, line 21:
```suggestion
        name: containers-logs.tar.gz
```

should help people in understanding what logs they will find there

docs/README.md
```diff
@@ -337,6 +337,14 @@ Configures the git username and email to associate commits with the provided ide
...
 The two vars in the previous snippet are [workflow configuration variables](https://github.blog/changelog/2023-01-10-github-actions-support-for-configuration-variables-in-workflows/) that can be created at organization level and shared across different repositories.
 
+### docker-dump-containers-logs
+
+Dumps Docker containers logs. Every container's log will be stored in separate `<container_name>.log` file. All files will be archived under `logs.tar.gz` and available to download from workflow's summary page.
+
+```yaml
+      - uses: Alfresco/alfresco-build-tools/.github/actions/docker-dump-containers-logs@ref
+```
+
```
gionn, line 347: there is the toc at the start of the file to be updated (if you use vscode it gets automatically updated)

.github/actions/docker-dump-containers-logs/action.yml
```diff
@@ -0,0 +1,21 @@
...
+        docker ps -a --format '{{.ID}} {{.Names}}' | while read -r id name ; do
+          docker logs "${id}" --tail 5000 > "${name}.log"
+        done
+        # Create archive with log files
+        tar -zcf logs.tar.gz *.log
+    - name: "Upload archive containing all *.log files"
+      uses: actions/upload-artifact@v3
+      with:
+        path: logs.tar.gz
+        name: containers-logs.tar.gz
```
dsibilio, line 21: In retrospective, if we wish to use this action in several jobs/steps within the same workflow, giving it a static name will make it so that the last artifact wins and overrides the previous one as far as I understand.
Shouldn't we add some uniqueness to this artifact's name so that it can be used more and with less preoccupations?

@gionn not sure if you have any suggestions regarding the name uniqueness, some that come to mind:
- time-based _(e.g.: including the evaluated `$(date +%s)` in the artifact name)_
- random-based _(.e.g.: including a UUID in the artifact name)_
- input-based _(governed by an input in this action, so that the users can easily identify which artifact is linked to which step/job as the final artifact name would be something like `${INPUT_IDENTIFIER}-containers-logs.tar.gz`)_

Ideally it could also be a mix and match of these solutions, where if the user does not specify any input than a pseudo-random unique name will be used _(e.g. via the time-based approach)_, whereas if the user does specify the input then that will be preferred. So we keep this super easy to use as there is no required configuration, but at the same time we make it easier to use efficiently in bigger workflows.

dsibilio (changes requested):

.github/actions/docker-dump-containers-logs/action.yml
krdabrowski: Good point Domenico! I'm in favor of mixed solution - to allow users to specify the output archive name entirely, and if not specified then use the pseudo random one. I would also consider evaluating `$(date +%Y%m%d%H%M%S)` as it's more meaningful then epoch seconds.

gionn: `${ github.job }` which reference the job id and `${ github.run_attempt }` which contains the retry number (starting from 1)?

dsibilio (dismissed):
LGTM, assuming it's been tested and works as usual 👍 

Even if the logic of the `job_id` + `run_attempt` is very solid to guarantee uniqueness across a workflow, I think it's still a good decision to allow users to rename the artifact if they want to since it can cover more scenarios which could potentially be problematic with the default method _(e.g.: I believe a matrix would cause the final artifact to have the same name multiple times, unless the user specifies the appropriate name for the artifact depending on the specific matrix execution)_.

Left a minor suggestion to fix up the docs.

.github/actions/docker-dump-containers-logs/action.yml
krdabrowski: I'm testing the solution with job id and retry number, but I'm afraid it won't be applicable in case of usage in multiple steps within a job. Adding timestamp might be necessary.

krdabrowski:
I've tested it and used this action in 2 steps within one job. In result 2 files were created: `containers-logs-arm64_health_check-1-20230531094611.tar.gz` and `containers-logs-arm64_health_check-1-20230531094615.tar.gz`.  After re-run attempts counter increased. Test workflow: https://github.com/Alfresco/acs-packaging/actions/runs/5131238912

dsibilio (approved):

gionn (approved):

PR #267, opened 2023-05-30
AFI-1518: Introduce action for Microsoft Teams notifications

rahulhyland:
Introduction of the Microsoft Teams notification for Github actions. After reviewing the Github marketplace, I found the [Notify Microsoft Teams](https://github.com/marketplace/actions/notify-microsoft-teams) is currently the most active and used. Utilizing the existing library Notify Microsoft Teams from the marketplace we're introducing our own action which can be consumed by other repos if there is a need.


docs/README.md
```diff
@@ -1034,6 +1035,71 @@ This message can also be appended to the default message:
...
+![Teams Success](./images/teams-success.png)
+
+Sample of a FAILURE notification on a `push` event.
+
+![Teams Failure](./images/teams-failure.png)
+
+Below is the detailed description of the above message card.
+
+- The `summary` of the message card tells you the author, event name, and the repo where it was performed.
+- The below `change log` will fetch the commit details.
```
gionn, line 1061:
```suggestion
- The below `Changelog` section will list the commits.
```

docs/README.md
```diff
@@ -1034,6 +1035,71 @@ This message can also be appended to the default message:
...
+- The `build_and_test` and `publish` are the job names added as a `need`. If you want to display needs status, you need to explicitly send the JSON object of the `needs` context as an input.
+
+  ```yml
+      uses: Alfresco/alfresco-build-tools/.github/actions/send-teams-notification@ref
+      with:
+        webhook-url: ${{ secrets.MSTEAMS_WEBHOOK }}
+        needs: ${{ toJson(needs) }}
```
gionn, line 1068: I am not getting this `needs` variable where come from, it's not a default one?

docs/README.md
```diff
@@ -1034,6 +1035,71 @@ This message can also be appended to the default message:
...
+  ```
+
+  The `Failure` status with an image reflects the workflow status.
+  For the ease of access, these action buttons will perform the following:
+
+  - Repository: It will open the Github Repository where the action took place.
+  - Build URL: It will redirect you to the workflow run under Github actions tab.
+  - Compare: It will redirect you to the Github GUI to review the changes for these commits.
```
gionn, line 1076: you are being very detailed, however what those buttons do is pretty self-explanatory, I think we can avoid documenting that and just focus on how to use this action

.github/actions/send-teams-notification/action.yml
```diff
@@ -0,0 +1,72 @@
...
+                ]
+              }
+            ]
+      }
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@v3
+    - uses: martialonline/workflow-status@v4
+      id: workflow_status_check
```
gionn, line 63: better add a name to this step to explain which is the purpose of that, I am not getting it

gionn (changes requested):

gionn:
and please use the subject to describe which is the purpose of the PR in addition to the ticket ref

docs/README.md
rahulhyland: Yes, its the [needs-context](https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context) object provided by Github itself. You need to pass it from the workflow itself as its not available with in the action. Its an optional `input` that will show the status of the jobs at a glance in a workflow in the Teams message card.

docs/README.md
rahulhyland: For a moment, I also thought the same. I'll get rid of it.

.github/actions/send-teams-notification/action.yml
rahulhyland: I have added a name against the step.

docs/README.md
gionn: interesting, it's the first time I see it used 👍🏻

gionn (approved):

docs/README.md
```diff
@@ -1034,6 +1035,67 @@ This message can also be appended to the default message:
...
+
+  **Details of input parameter of this action:**
+
+  | Input     | Description                                                                       |
+  |-----------------------------------------------------------------------------------|------------|
+  | webhook-url | URL of the MS Teams incoming webhook.                                             |
+  | needs | JSON parsed needs context.                                                        |
+  | dry-run | Do not actually send the message card to the teams channel.                       |
+  | raw | The entire JSON object of the Message Card which will be sent to Microsoft Teams. |
+  | overwrite | JSON like object to overwrite default message.                                    |
```
dsibilio, line 1097:
```suggestion
  | Input       | Description                                                                       |
  |-------------|-----------------------------------------------------------------------------------|
  | webhook-url | URL of the MS Teams incoming webhook.                                             |
  | needs       | JSON parsed needs context.                                                        |
  | dry-run     | Do not actually send the message card to the teams channel.                       |
  | raw         | The entire JSON object of the Message Card which will be sent to Microsoft Teams. |
  | overwrite   | JSON like object to overwrite default message.                                    |
```

dsibilio: _(markdown formatting)_

dsibilio (dismissed):
LGTM. Left some formatting suggestion for the README markdown table.

alxgomz (approved):

rahulhyland:
@gionn , @dsibilio , @alxgomz , please approve it once again as the commit for suggested README changes staled the review.

gionn (approved):

dsibilio (approved):

PR #270, opened 2023-06-07
AAE-13340 enable testcontainers reuse

jsokolowskii:
Add option to enable reuse of testcontainers to speed up build process.
Documented feature: https://alfresco.atlassian.net/wiki/spaces/HXP/pages/1275265126/PoC+Parallel+Testing+Multi-Thread+Maven+Build#TestContainers-Reuse

### DIFF
--- a/.github/actions/maven-build-and-tag/action.yml
+++ b/.github/actions/maven-build-and-tag/action.yml
@@ -156,7 +156,9 @@ runs:
     - name: Enable testcontainers reuse option
       if: inputs.reuse-testcontainers == 'true'
       shell: bash
-      run: echo "testcontainers.reuse.enable=true" > ~/.testcontainers.properties
+      run: | 
+        echo "testcontainers.reuse.enable=true" > ~/.testcontainers.properties
+        echo "TESTCONTAINERS_RYUK_DISABLED=true" >> $GITHUB_ENV
 
     - name: Build and Test with Maven (and maybe Deploy)
       shell: bash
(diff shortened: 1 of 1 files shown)


andrea-ligios (approved):
LGTM!

atchertchian (approved):

erdemedeiros (approved):

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -157,6 +168,11 @@ runs:
         MAVEN_USERNAME: ${{ inputs.maven-username }}
         MAVEN_PASSWORD: ${{ inputs.maven-password }}
 
+    - name: Remove running docker containers
+      if: inputs.reuse-testcontainers == 'true'
+      shell: bash
+      run: docker rm -f $(docker ps -a -q)
```
atchertchian, line 174: Quick note: I'm wondering about potential termination of containers that would not have been started by testcontainers, but maybe we don't have such use cases.
It seems that there's no easy way to isolate them anyway (unless we added specific labels in tests for each container...) according to https://github.com/testcontainers/testcontainers-java/issues/3569

atchertchian (approved):

.github/actions/maven-build-and-tag/action.yml
jsokolowskii: Thank you @atchertchian! Yes, I was also wondering about that, but AFAIK we don't have yet such use-cases. It would be necessary if we had though - something to consider in the future.

PR #272, opened 2023-06-12
[AAE-15185] Add step of uploading coverage report in maven-build-and-tag

mavotto:
as the title says in this PR it was added the upload of the coverage report if required.

https://alfresco.atlassian.net/browse/AAE-15185

### DIFF
--- a/.github/actions/maven-build-and-tag/action.yml
+++ b/.github/actions/maven-build-and-tag/action.yml
@@ -171,7 +171,7 @@ runs:
         name: coverage
         retention-days: 1
         path: |
-          **/target/site/jacoco-aggregate/*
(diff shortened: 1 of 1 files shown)


.github/actions/maven-build-and-tag/action.yml
```diff
@@ -161,6 +165,14 @@ runs:
         path: |
           **/target/*.jar
 
+    - uses: actions/upload-artifact@v3
+      if: inputs.upload-coverage == 'true'
+      with:
+        name: coverage
+        retention-days: 1
```
gionn, line 172: isn't that too low? I would expect to keep it at least for a week in case you want to compare previous reports?

gionn (approved):

PR #277, opened 2023-06-21
OPSEXP-2106 add new action pipenv

slohe1:


.github/actions/pipenv/requirements.txt
```diff
@@ -0,0 +1 @@
+pipenv==2023.6.12
```
gionn, line 1: there is also a dependabot configuration that needs to be moved in order to keep this updated

.github/actions/pipenv/action.yml
```diff
@@ -0,0 +1,12 @@
...
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install pipenv dependency
+      shell: bash
+      run: pip install -r ${{ github.action_path }}/requirements.txt
+    - name: Prepare pipenv virtualenv
+      shell: bash
+      run: pipenv install --deploy --dev --python $(cat .python-version)
```
gionn, line 12: remember to always add a readme entry for each new action

document also that the .python-version file in the repo root is a requirement (for now)

gionn (changes requested):

alxgomz (commented):
Shouldn't this action setup/python as well?

gionn:
> Shouldn't this action setup/python as well?

I prefer to avoid it and push on composability in the workflow, otherwise we tend to have too much restrictive actions and little added utility overall

gionn (changes requested):

docs/README.md
```diff
@@ -844,6 +845,21 @@ Or:
           check-github-configuration: 'false'
 ```
 
+### pipenv
+
+This workflow sets up a Python environment with version 3.9 using the actions/setup-python@v4 action and utilizes the pipenv action to manage Python dependencies based on the specified Python version.
```
gionn, line 850:
```suggestion
This workflow sets up a Python environment using the standard setup-python action and utilizes the pipenv action to manage Python dependencies declared in the Pipfile and based on the specified Python version
```

gionn (approved):

PR #282, opened 2023-06-26
PRODENG-222 Action to run PMD scan.

tpage-alfresco:


tpage-alfresco:
For now we have the first draft of the default ruleset defined here: https://github.com/AlfrescoLabs/pmd-ruleset/

.github/actions/pmd/action.yml
```diff
@@ -0,0 +1,46 @@
...
+    default: "true"
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/checkout@v3
+      env:
+        REF_TO_CHECKOUT: ${{ inputs.auto-commit == 'true' && github.head_ref || '' }}
+      with:
+        ref: ${{ env.REF_TO_CHECKOUT }}
```
dsibilio, line 27: I'm not quite sure what's this section here for? I don't see any `auto-commit` inputs in this action.

.github/actions/pmd/action.yml
```diff
@@ -0,0 +1,46 @@
...
+
+    - name: Download ruleset
+      run: curl https://raw.githubusercontent.com/AlfrescoLabs/pmd-ruleset/master/pmd-ruleset.xml -o ${{ github.action_path }}/pmd-ruleset.xml
```
dsibilio, line 30: I'd assume it would be ideal to provide a configurable input to customize the branch we're pulling the ruleset from, in case we need to handle different rulesets depending on the project?

dsibilio (changes requested):
Reminder to also update the https://github.com/Alfresco/alfresco-build-tools#github-actions README file with an example of this action's usage + adding it to the ToC.

.github/actions/pmd/action.yml
tpage-alfresco: Ultimately I thought we would get rid of the external dependency and move the default ruleset(s) into this project. Alternatively we could make it configurable as you suggest.

dsibilio: I was thinking about the 2 approaches as well, and while in most cases we tend to aggregate all necessary configuration files here, I think that this could be a reasonable exception as **it makes it so that we always pull the latest rulesets no matter which version of alfresco-build-tool we're using**. Furthermore I think we can slightly better leverage git to handle multiple rulesets as separate branches in a separate repository, than multiple files in the same branch _(the latter would require more manual copy-pasting in case of global updates as opposed to just a set of pulling from master + solving eventual conflicts, at the same time the "knowledge gate" to update all rulesets would be definitely lower than having to acknowledge the existence of a separate repository with multiple branches in the very same scenario -- so there's this tradeoff)_.

I'm slightly in favor of keeping the rulesets in a separate repository if the general consensus is that we'd work slightly better with 1 single ruleset file per branch over multiple branches + we could really benefit from being able to skip upgrading N repositories PMD action's version just to pull a newer version of a ruleset. Ultimately if you can think of reasons why we'd work better a different way I'd be glad to change my opinion though :)

gionn: you can alse have another `actions/checkout@v3` step for the ruleset specific repo, adding an input to change the ref (likely to what domenico is suggesting)

I think we will figure it out once we integrate it in more projects which is the best way to handle rulesets, for now the current external repo approach can be enough for me

.github/actions/pmd/action.yml
gionn: not sure either what env.REF_TO_CHECKOUT is for, usually it should be smart enough without parameters

.github/actions/pmd/delta-scan.sh
```diff
@@ -0,0 +1,65 @@
...
+    if [[ -f ${file} ]]
+    then
+        echo ${file} > ${tmp_dir}/old-files.txt
+    fi
+done
+${run_pmd} pmd --cache ${tmp_dir}/pmd.cache --file-list ${tmp_dir}/old-files.txt -R ${ACTION_PATH}/pmd-ruleset.xml -r ${tmp_dir}/old_report.txt --fail-on-violation false
+old_issue_count=$(cat ${tmp_dir}/old_report.txt | wc -l)
+
+# Rerun PMD against the PR head commit.
+git checkout ${head_ref}
```
gionn, line 35: I think you can simplify this script, when you need to checkout always use an external step `actions/checkout@v3` and then just call this (simplified) script with different parameters each time which just invoke pmd once

.github/actions/pmd/action.yml
tpage-alfresco: It was probably in whatever action I copied to get started! I'll simplify this to the defaults.

dsibilio: Usually we don't even include the checkout step in our actions at all, it's assumed that the user checks out the ref they're interested in. I think you might have copied this from an action which has an edge-case such as it wants to commit things on a specific branch in case of auto-commit being true (e.g.: the pre-commit action).

Personally I'd get rid of the whole step.

.github/actions/pmd/delta-scan.sh
tpage-alfresco: I think that makes it harder to follow the behaviour really. As a really minor consideration it also makes the action less portable to other CI systems. If we extract the common code like this we would end up with something like:
* Run a script to find `baseline_ref` and to create `file-list.txt`
* Run checkout@v3
* Run the PMD script against the old commit, storing the results in `old_report.txt`
* Run checkout@v3
* Run the PMD script against the new commit, storing the results in `new_report.txt`
* Run a diff script to compare the violations and potentially fail the build.

I did consider extracting the common code (checkout, run PMD, count issues) into a bash function, but I didn't think it was worthwhile.

.github/actions/pmd/action.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Install PMD
+      uses: pmd/pmd-github-action@d57c0463ebba262a33d1983a5c6ac6031bfde43b # v1.4.1
+      id: pmd
+      with:
+        version: ${{ inputs.pmd-version }}
+        rulesets: 'pmd-ruleset.xml'
+        createGitHubAnnotations: ${{ inputs.create-github-annotations }}
+
+    - name: Run PMD scan against changes
+      run: ${{ github.action_path }}/delta-scan.sh "origin/${{ github.base_ref }}" "${{ github.sha }}" "${{ inputs.fail-on-new-issues }}"
```
tpage-alfresco, line 41: This should only be run against PRs. Need to add something to ensure this is skipped for other triggers.

gionn (approved):

dsibilio (approved):

jamalkm (approved):
LGTM. Thanks Tom for sorting this out.

PR #300, opened 2023-08-02
Adding ADDITIONAL_SETTINGS usage to veracode scan job. Pipeline scans…

mstrankowski:
… that include test modules, which aren't present in release artifacts, often show up in SCA scans of components creating a false sense of vulnerability, that isn't there. Test modules should be excluded from releases and from security scans, but no other way than directly amending this command would work

### DIFF
--- a/.github/actions/veracode/source_clear.sh
+++ b/.github/actions/veracode/source_clear.sh
@@ -9,7 +9,7 @@ mvn -B -q clean install \
     -Dmaven.javadoc.skip=true \
     com.srcclr:srcclr-maven-plugin:scan \
     -Dcom.srcclr.apiToken=${SRCCLR_API_TOKEN} > scan.log \
-    $ADDITIONAL_SETTINGS
+    ${ADDITIONAL_SETTINGS}
 
 SUCCESS=$?   # this will read exit code of the previous command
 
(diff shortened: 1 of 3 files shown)


.github/actions/veracode/source_clear.sh
```diff
@@ -8,7 +8,8 @@ mvn -B -q clean install \
     -DskipTests \
     -Dmaven.javadoc.skip=true \
     com.srcclr:srcclr-maven-plugin:scan \
-    -Dcom.srcclr.apiToken=${SRCCLR_API_TOKEN} > scan.log
+    -Dcom.srcclr.apiToken=${SRCCLR_API_TOKEN} > scan.log \
+    ${ADDITIONAL_SETTINGS}
```
gionn, line 12: better leverage the action input and not rely on an arbitrary env var, it could break everything if a build define ADDITIONAL_SETTINGS

at this point to achieve maximum flexibility, I would define two inputs:
* `srcclr_install_default_options` which defaults to `-DskipTests -Dmaven.javadoc.skip=true`
* `srcclr_install_default_options` which defaults to empty and may be set to provide additional parameters in downstream projects

gionn (changes requested):

gionn (dismissed):

gionn (approved):

PR #309, opened 2023-08-07
AAE-15944: allow specifying nginx manifest in setup-kind

atchertchian:
### Checklist

- Jira Reference (also in PR title): AAE-15944
- [README](/docs/README.md) updated after adding/changing behaviour of an action
- Proposed version increment for [release](/docs/README.md#release):
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)


.github/actions/setup-kind/action.yml
```diff
@@ -19,11 +23,8 @@ runs:
 
     - name: Install ingress-nginx
       shell: bash
-      env:
-        NGINX_MANIFEST: |
-          https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml
       run: |
-        kubectl apply -f $NGINX_MANIFEST
+        kubectl apply -f ${{ inputs.nginx-manifest }}
```
gionn, line 27: how do you plan to use it? (link a PR)

I think there are very few situations where we need to completely provide a custom nginx install manifest, while the most common use case will be to install a specific ingress-nginx version - so providing an input to customize the current `/main/` subpath should be enough?

atchertchian: I did not validate if changing the version would be enough: I'll set that work aside for now then

alxgomz: I actually just faced a use case where that would have been useful as a setting has been changed one hour ago and it would have been useful to be able to pin the version

alxgomz: To add on top of Gio's comment I would add that if changing the version would not be enough one can still patch the deployment manually as he needs to in a different tasks. If you don't min I'll take over this PR and amend to turn `nginx-manifest` input to a `nginx-igress-ref`?

atchertchian: This PR is just a draft but feel free to reuse it: as I said above I did not validate the changes for which this would have been useful.
Also as discussed with @gionn the input does not need to be the whole URL, it could be controlling the "main" part in the URL only

atchertchian: yes @alxgomz please go ahead (and FYI my dependent unsuccessful PR is/was https://github.com/Activiti/activiti-cloud/pull/1165)

atchertchian:
done with #326 --> resolve

PR #314, opened 2023-09-01
AAE-16243 add ignore empty commit option

jsokolowskii:
### Checklist

- Jira Reference (also in PR title): https://alfresco.atlassian.net/browse/AAE-16243
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)
- External PR link where changes has been tested: https://github.com/Alfresco/alfresco-process-releases/pull/434

### Description
Adds an option to ignore commit which would be empty for the `git-commit-changes` action.


dsibilio (commented):
Ultimately I'm not quite sure I understand the issue, in order to commit successfully with no changes the consumer would already need to pass `inputs.commit-options='--allow-empty'`.

I'm just trying to understand, why is all this logic needed instead of the consumer simply not passing `--allow-empty`? If it's simply because of the git commit exit code turning into `1` if the `--allow-empty` is not provided and nothing had been staged for commit?

Also, does anyone realistically want empty commits? It seems to me that instead of being configurable it could just be turned off

.github/actions/git-commit-changes/action.yml
```diff
@@ -17,19 +17,40 @@ inputs:
...
+    default: 'false'
 runs:
   using: composite
   steps:
-    - name: Commit stagged changes
+    - name: Configure git user
+      shell: bash
+      run: |
+        git config --global user.name ${{ inputs.username }}
+        git config --global user.email ${{ inputs.username }}@users.noreply.github.com
```
gionn, line 31: there is already an action for this
https://github.com/Alfresco/alfresco-build-tools#configure-git-author

in general it's better to not embed an action inside another action, just use the existing action before this action in your workflow

gionn (commented):
I didn't remember we had this git-commit-changes action here, I would consider switching to https://github.com/stefanzweifel/git-auto-commit-action which is well maintained and we are already using it in different repos instead of trying to reinvent the wheel here

erdemedeiros:
> I didn't remember we had this git-commit-changes action here, I would consider switching to https://github.com/stefanzweifel/git-auto-commit-action which is well maintained and we are already using it in different repos instead of trying to reinvent the wheel here

I agree with you that's better to use what is already available instead of recreating it. However, back in the time I remember that the policy about which kind of external actions could be used or not was not well defined. There were some security concerns IIRC. For instance, the action you mentioned does not have the label `Verified creator`. But maybe we have some guidelines on this now and I guess if it's already used in some places it should be fine to use it in other places as well.

jsokolowskii:
> Ultimately I'm not quite sure I understand the issue, in order to commit successfully with no changes the consumer would already need to pass `inputs.commit-options='--allow-empty'`.
> 
> I'm just trying to understand, why is all this logic needed instead of the consumer simply not passing `--allow-empty`? Is it simply because of the git commit exit code turning into `1` if the `--allow-empty` is not provided and nothing had been staged for commit?
> 
> Also, does anyone realistically want empty commits? It seems to me that instead of being configurable it could just be turned off

The idea here was to allow empty commits if the option is provided and also make it possible to ignore the commit if the option is not given and we don't want the action to fail. If everyone agrees to return 0 if the commit would be empty I'm ok with that also. I don't think `allow-empty` is used anywhere other than alfresco-process-releases.

.github/actions/git-commit-changes/action.yml
```diff
@@ -17,19 +17,40 @@ inputs:
...
+
+    - name: Check for files to Commit
+      id: any-files-to-commit
+      shell: bash
+      working-directory: ${{ matrix.repository }}
```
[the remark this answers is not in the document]
jsokolowskii: Sorry, I copied it from elsewhere - I'll remove it

jsokolowskii:
> I didn't remember we had this git-commit-changes action here, I would consider switching to https://github.com/stefanzweifel/git-auto-commit-action which is well maintained and we are already using it in different repos instead of trying to reinvent the wheel here

I'm not sure if we can accomplish the same by using the referenced action - we would still need to wrap it up in our logic. Maybe it would be good to use it where possible though.

.github/actions/git-commit-changes/action.yml
erdemedeiros: Yes, this generally true, but in some cases it makes sense to group together several actions that are always called together, specially when they are called in several places.
Note that the current action is already like that, these two lines were just moved up. Given the fact that this action is used in a lot of places, it's not viable to split them in the callers now. However, maybe the two lines can be replaced by a call to `configure-git-author`.

PR #380, opened 2023-11-21
AAE-17877 m2 cache should be only restored on build startup

wojciech-piotrowiak:
### Checklist

- Jira Reference (also in PR title): AAE-17877
  - [ ] Patch (bugfix)
  - [X] Minor (new feature)


.github/actions/setup-java-build/action.yml
```diff
@@ -16,8 +16,8 @@ inputs:
 runs:
   using: composite
   steps:
-    - name: "Cache local Maven repository"
-      uses: actions/cache@v3
+    - name: "Restore cache local Maven repository"
+      uses: actions/cache/restore@704facf57e6136b1bc63b828d79edcd491f0ee84 # v3.3.2
```
gionn, line 20: no one is going to save the cache if you move to restore here

gionn (changes requested):
you need to better describe what you are trying to achieve here because I am completely missing the point

.github/actions/setup-java-build/action.yml
atchertchian: also this is the setup-java action: we're not using it in most of our repos: I think this change is not needed for the task at hand

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -81,13 +81,6 @@ runs:
   steps:
     - uses: actions/checkout@v4
 
-    - uses: actions/cache@v3
```
atchertchian, line 84: I think this should be kept, just made dependent on whether we're handling a PR or a push: if it's a push we keep things as is, if it's a PR we only restore cache with the given restore keys, instead of also saving it at the end of the build

wojciech-piotrowiak: this entry makes no sense here, because the same command is located in setup-jva

atchertchian: we're not calling the local `./actions/setup-java-build` here, but the regular `actions/setup-java`, and the latter should not handle cache at all, that was the purpose of https://github.com/Alfresco/alfresco-build-tools/pull/147

wojciech-piotrowiak: 🤦‍♂️ 
Ok, that changes a lot ! sorry for that

atchertchian: no worries :+1:

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -81,12 +81,10 @@ runs:
   steps:
     - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
-    - uses: actions/cache@704facf57e6136b1bc63b828d79edcd491f0ee84 # v3.3.2
+    - uses: actions/cache/restore@704facf57e6136b1bc63b828d79edcd491f0ee84 # v3.3.2
       with:
         path: ~/.m2/repository
         key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
-        restore-keys: |
-          ${{ runner.os }}-maven-
```
atchertchian, line 89: These could still be useful, no?

version.txt
```diff
@@ -1 +1 @@
-v5.0.0
+v5.1.0
```
atchertchian, line 1:
```suggestion
v5.2.0
```
(v5.1.0 was released in the meantime)

.github/actions/maven-build-and-tag/action.yml
wojciech-piotrowiak: right, thanks

atchertchian (approved):

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -229,3 +229,17 @@ runs:
...
+    - name: Clean m2 cache
+      if: github.event_name == 'push'
+      shell: bash
+      run: |
+        rm -fr ~/.m2/repository/org/activiti
+        rm -fr ~/.m2/repository/com/alfresco
+
+    - name: Save maven cache
+      uses: actions/cache/save@704facf57e6136b1bc63b828d79edcd491f0ee84 # v3.3.2
+      if: github.event_name == 'push'
```
gionn, line 242: this doesn't prevent saving cache for branches different than the default one if the caller workflow is configured to build also those

something like `&& github.event.repository.default_branch == github.ref_name` should fix that

gionn (approved):

.github/actions/maven-build-and-tag/action.yml
wojciech-piotrowiak: thanks!

atchertchian: Actually I think we want to keep the cache on maintenance branches, that's why keeping the logic checking event `push` makes more sense to me

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -229,3 +229,17 @@ runs:
       run: |
         git tag -a $VERSION -m "Release version $VERSION"
         git push origin $VERSION
+
+    - name: Clean m2 cache
+      if: github.event_name == 'push'
```
atchertchian, line 234: also if the condition is changed this should be kept consistent

gionn (approved):

.github/actions/maven-build-and-tag/action.yml
erdemedeiros: Yes, I think we need to consider maintenance branches as well, and we usually don't configure the CI to build on feature branches, so it should not be a problem.

erdemedeiros (approved):
LGTM!

atchertchian (approved):

PR #382, opened 2023-11-22
AAE-18225 Use SHA for GitHub actions

Giovanni007:
### Checklist

  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)


.github/workflows/test.yml
```diff
@@ -7,11 +7,16 @@ on:
...
 
 jobs:
-
+  pre-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Ensure SHA pinned actions
+        uses: zgosalvez/github-actions-ensure-sha-pinned-actions@f32435541e24cd6a4700a7f52bb2ec59e80603b1 # v2.1.4
   test:
```
Giovanni007, line 16: Not sure if I should add this pre-check here or maybe in the `release` workflow. I just need to run the Ensure SHA pinned actions check.  I added it in a separated job just to separate the concern .
Please advise if there is a better location for it.
cc: @gionn

wojciech-piotrowiak (approved):

.github/workflows/test.yml
gionn: looks good here, otherwise you get a failure once the pr is already merged

can you rename the new job to a more generic `build` and move here the `- uses: ./.github/actions/pre-commit` step below?

version.txt
```diff
@@ -1 +1 @@
-v3.8.1
+v3.8.2
```
gionn, line 1: I know technically we are not changing anything in the behavior of che actions but since we are doing quite a big change across all the actions we could actually ship it as v4.0.0

gionn (changes requested):

.github/actions/dbp-charts/verify-helm/action.yml
```diff
@@ -60,26 +60,26 @@ runs:
...
       with:
         fetch-depth: 0
-    - uses: aws-actions/configure-aws-credentials@v4
+    - uses: aws-actions/configure-aws-credentials@010d0da01d0b5a38af31e9c3470dbfdabdecca3a # v4.0.1
       with:
         aws-region: ${{ inputs.aws_region }}
         aws-access-key-id: ${{ inputs.aws_access_key_id }}
         aws-secret-access-key: ${{ inputs.aws_secret_access_key }}
-    - uses: azure/setup-kubectl@v3
+    - uses: azure/setup-kubectl@901a10e89ea615cf61f57ac05cecdf23e7de06d8 # v3
```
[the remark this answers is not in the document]
Giovanni007: I think that's because they don't use semver versioning

gionn (approved):

PR #449, opened 2024-01-11
AAE-19391: improve GH run URL in notifications, when rerunning

atchertchian:
### Checklist

- Jira Reference (also in PR title): AAE-19391
  - [ ] Patch (bugfix)
  - [X] Minor (new feature)
- External PR link where changes has been tested: see mention

### Description

Slack notification now gives a link to the given attempt, in case of rerun (and also states so in the workflow title)


.github/actions/send-slack-notification/action.yml
```diff
@@ -81,6 +81,26 @@ runs:
...
+        echo "result=$RESULT" >> $GITHUB_OUTPUT
+
+    - name: Compute workflow title
+      id: compute-workflow-title
+      shell: bash
+      run: |
+        RESULT="$GITHUB_WORKFLOW"
+        if [[ "$GITHUB_RUN_ATTEMPT" != '1' ]]; then
+          RESULT+=" (attempt #$GITHUB_RUN_ATTEMPT)"
+        fi
```
gionn, line 101: that means that the first run attempt will always link to the last attempt, right?

gionn (approved):

.github/actions/send-slack-notification/action.yml
atchertchian: No the URL will always have attempt/1, here it's just the title that does not state "(attempt 1)" after the workflow title

PR #498, opened 2024-02-09
AAE-20018 Add whitelist-branch parameter in enforce-pr-conventions

RahulKumar1606:
### Checklist

  - [x] Patch (bugfix)
  - [ ] Minor (new feature)

### Description

Add whitelist-branch parameter in enforce-pr-conventions

### DIFF
--- a/.github/actions/enforce-pr-conventions/action.yml
+++ b/.github/actions/enforce-pr-conventions/action.yml
@@ -12,6 +12,9 @@ inputs:
   valid-pr-title-regex:
     description: The regular expression to check the pull request title
     required: false
+  whitelist-branches:
+    description: The list of branches that are allowed to bypass the branch check
+    required: false
 
 runs:
   using: composite
@@ -61,7 +64,7 @@ runs:
         fi
 
     - name: Check branch name
-      if: steps.check.outputs.continue == 'true'
+      if: steps.check.outputs.continue == 'true' && !contains(inputs.whitelist-branches, env.BRANCH_NAME)
       shell: bash
       env:
         LOCAL_BRANCH: ${{ env.BRANCH_NAME }}
--- a/docs/README.md
+++ b/docs/README.md
@@ -430,7 +430,14 @@ If default regular expressions do not match the need, they can also be defined:
           valid-branch-regex: "^(feature|test|tmp)\/JKEY-[0-9]+-[A-Za-z0-9._-]+$"
           valid-pr-title-regex: "^JKEY-[0-9]+ [A-Za-z]{1}.*$"
 ```
+To exempt specific branch names from the branch name checks, the optional input parameter called `whitelist-branch` can be utilized. If there are multiple branches to be excluded, they can be separate using the "|" symbol.
 
+```yaml
+      - uses: Alfresco/alfresco-build-tools/.github/actions/enforce-pr-conventions@ref
+        with:
+          jira-project-key: JKEY
+          whitelist-branch: "ABC-1234-branch|XYZ-5678-branch"
+``` 
 The inputs `jira-project-key`, `valid-branch-regex` and `valid-pr-title-regex` are optional: if `valid-branch-regex` or `valid-pr-title-regex` are not provided, the action will consume `jira-project-key` to generate the default regex.
 
 **Default regex for Branch name**: `"^(improvement|bug|feature|test|tmp)/(<jira-project-key>)-[0-9]+-[A-Za-z0-9._-]+$"`


docs/README.md
```diff
@@ -431,6 +431,15 @@ If default regular expressions do not match the need, they can also be defined:
           valid-pr-title-regex: "^JKEY-[0-9]+ [A-Za-z]{1}.*$"
 ```
 
+To exempt specific branch names from the branch name checks, the optional input parameter called `whitelist-branch` can be utilized. If there are multiple branches to be excluded, they can be separate using the "|" symbol.
+
+```yaml
+      - uses: Alfresco/alfresco-build-tools/.github/actions/enforce-pr-conventions@ref
+        with:
+          jira-project-key: JKEY
+          whitelist-branch: "ABC-1234-branch|XYZ-5678-branch"
```
andrea-ligios, line 440: should be `whitelist-branches`

andrea-ligios (changes requested):

docs/README.md
```diff
@@ -431,6 +431,15 @@ If default regular expressions do not match the need, they can also be defined:
           valid-pr-title-regex: "^JKEY-[0-9]+ [A-Za-z]{1}.*$"
 ```
 
+To exempt specific branch names from the branch name checks, the optional input parameter called `whitelist-branch` can be utilized. If there are multiple branches to be excluded, they can be separate using the "|" symbol.
```
andrea-ligios, line 434: should be `whitelist-branches`

andrea-ligios (dismissed):

docs/README.md
```diff
@@ -431,6 +431,15 @@ If default regular expressions do not match the need, they can also be defined:
           valid-pr-title-regex: "^JKEY-[0-9]+ [A-Za-z]{1}.*$"
 ```
 
+To exempt specific branch names from the branch name checks, the optional input parameter called `whitelist-branches` can be utilized. If there are multiple branches to be excluded, they can be separated using the "|" symbol.
+
+```yaml
+      - uses: Alfresco/alfresco-build-tools/.github/actions/enforce-pr-conventions@ref
+        with:
+          jira-project-key: JKEY
+          whitelist-branches: "ABC-1234-branch|XYZ-5678-branch"
```
atchertchian, line 440: I don't see why that syntax would work natively with contains, did you try this on a test branch please?
if that helps: https://stackoverflow.com/questions/75420197/how-to-use-array-input-for-a-custom-github-actions

Also there are pre-commit errors on that file

RahulKumar1606: Actually I found that we can use `contains` to check if a `current branch` is present in `whitelist-branches` or not, if present then we will not run the branch check.
More info can be found [here](https://docs.github.com/en/actions/learn-github-actions/expressions#contains).

There is one mistake that I found few minutes ago is what if someone's current branch name is `XYZ` and `whitelist-branches` is `ABC-1234-branch|XYZ-5678-branch`, `contains` will return true in this case as `whitelist-branches` has a substring with `XYZ`. So right now I am modifying it to convert to list and using for loop to check if current branch matches the lists of `whitelist-branches`

RahulKumar1606: Or shall we take json inputs as mentioned here -> https://stackoverflow.com/questions/75420197/how-to-use-array-input-for-a-custom-github-actions?

atchertchian: When I look at the answer I think it's better to do that, here's a quick test that worked:

with:
```
        with:
          whitelist-branches: |-
            foo
            bar
```

some bash that did print "foo found": 
```
      run: |
        readarray -t array <<< "${{ inputs.whitelist-branches }}"
        if [[ ${array[@]} =~ 'foo' ]]
        then
          echo "foo found"
        else
          echo "foo not found"
        fi
```

.github/actions/enforce-pr-conventions/action.yml
```diff
@@ -60,8 +63,22 @@ runs:
...
+      shell: bash
+      run: |
+        readarray -t branches <<< "${{ inputs.whitelist-branches }}"
+        for branch in "${branches[@]}"; do
+            if [[ $branch == $BRANCH_NAME ]]; then
+                echo "continue=false" >> $GITHUB_OUTPUT
+                exit 0
+            fi
+        done
+        echo "continue=true" >> $GITHUB_OUTPUT
```
atchertchian, line 78: Would that work?

```suggestion
        if [[ ${branches[@]} =~ $BRANCH_NAME ]]; then
          echo "continue=false" >> $GITHUB_OUTPUT
        else
          echo "continue=true" >> $GITHUB_OUTPUT
        fi
```

RahulKumar1606: I guess it wont work because `=~` checks the regex pattern or maybe substring. I already tested this.
Eg if whitelisted-branch is `bug/AAE-123-conventional-commits` and if the current branch is `bug/AAE-123-conventional` then `if [[ ${branches[@]} =~ $BRANCH_NAME ]]` condition is true which is not correct.

atchertchian: Alright thanks :+1:

atchertchian (approved):
Thanks for this improvement :+1:

andrea-ligios (approved):

PR #507, opened 2024-02-20
AAE-20726 Add new action to fail on upcoming run detection

atchertchian:
### Checklist

- Jira Reference (also in PR title): AAE-20726
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)
- External PR link where changes has been tested: see mentions


.github/actions/github-check-upcoming-runs/action.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: Check upcoming runs
+      id: check
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
+        WORKFLOW: ${{ inputs.workflow }}
+      shell: bash
+      run: |
+        URL="/repos/$GITHUB_REPOSITORY/actions/workflows/$WORKFLOW/runs"
+        echo "Checking runs at $URL for $BRANCH_NAME"
+        RUNS=$(gh api -X GET $URL -F branch=$BRANCH_NAME -F event=pull_request)
```
gionn, line 31: maybe worth having event as input in case workflow is triggered via push

.github/actions/github-check-upcoming-runs/action.yml
```diff
@@ -0,0 +1,54 @@
...
+          echo $RUNS | jq --argjson id $GITHUB_RUN_ID '.workflow_runs[] | select(.id > $id) | {"id":.id, "status":.status, "conclusion":.conclusion}'
+
+          ID=$(echo $RUN | jq -r ".id")
+          STATUS=$(echo $RUN | jq -r ".status")
+          CONCLUSION=$(echo $RUN | jq -r ".conclusion")
+          RERUN_URL="$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$ID"
+
+          if [[ "$STATUS" == "completed" && "$CONCLUSION" == "cancelled" ]]; then
+            echo "::notice::Re-running $ID: $RERUN_URL"
+            gh run rerun -R $GITHUB_REPOSITORY $ID --failed
```
gionn, line 47: `--failed` is really needed? if the workflow has concurrency and is found cancelled, probably no job has ever run?

atchertchian: do you think it would fail if so? it's just to shortcut re-running every job that has gone through, but I can make this optional if you prefer, I do not expect it to fail if it was fully cancelled

.github/actions/github-check-upcoming-runs/action.yml
atchertchian: done

mavotto (approved):

gionn (approved):

PR #524, opened 2024-03-04
ACS-6793 Expand free disk space action

mikolajbrzezinski:
### Checklist

- Jira Reference (also in PR title): https://alfresco.atlassian.net/browse/ACS-6793
  - [ ] Patch (bugfix)
  - [X] Minor (new feature)
  
https://github.com/Alfresco/acs-packaging/pull/2698

### Description
For some tests the original action was not freeing up enough memory, so we decided to implement optional steps to help with that


.github/actions/free-hosted-runner-disk-space/action.yml
```diff
@@ -1,18 +1,148 @@
...
+        fi
+        if [[ ${{ inputs.remove-tools-cache }} != '' ]]; then
+          sudo rm -rf ${{ inputs.remove-tools-cache }}
+        fi
+        echo "... done"
+      shell: bash
+
+    - name: Maximize build space
+      if: ${{ inputs.only-remove-files == 'false' }}
+      uses: easimon/maximize-build-space@master
```
dsibilio, line 128: We can't use `master` as a reference here. We have to rely on a "tagged" version of the action, e.g.: via SHA pinning.

.github/actions/free-hosted-runner-disk-space/action.yml
```diff
@@ -1,18 +1,148 @@
...
+    default: 'true'
+  remove-docker-images:
+    description: 'Removes cached Docker images. (frees ~3 GB)'
+    required: false
+    default: 'false'
   to-remove:
-    description: Collection of files/folders to be removed
-    default: "/usr/share/dotnet /opt/ghc /usr/local/share/boost $AGENT_TOOLSDIRECTORY"
+    description: 'Additional collection of files/folders to be removed'
+    default: ""
```
dsibilio, line 72: I don't see this `to-remove` being used anywhere anymore.
I think it's useful to retain the ability to dynamically define more folders to delete, outside of the well-known ones.
IMO it should be used again, especially so that this won't be a breaking change upgrade!

.github/actions/free-hosted-runner-disk-space/action.yml
```diff
@@ -1,18 +1,148 @@
...
+    description: 'Swap space to create, in Megabytes.'
+    required: false
+    default: '1024'
+  overprovision-lvm:
+    description: |
+      Create the LVM disk images as sparse files, making the space required for the LVM image files *appear* unused on the
+      hosting volumes until actually allocated. Use with care, this can lead to surprising out-of-disk-space situations.
+      You should prefer adjusting root-reserve-mb/temp-reserve-mb over using this option.
+    required: false
+    default: 'false'
```
dsibilio, line 26: Are these used anywhere?

.github/actions/free-hosted-runner-disk-space/action.yml
```diff
@@ -1,18 +1,148 @@
...
+    required: false
+    default: 'runner:runner'
+  pv-loop-path:
+    description: 'Absolute file path for the LVM image created on the root filesystem, the default is usually fine.'
+    required: false
+    default: '/pv.img'
+  tmp-pv-loop-path:
+    description: 'Absolute file path for the LVM image created on the temp filesystem, the default is usually fine. Must reside on /mnt'
+    required: false
+    default: '/mnt/tmp-pv.img'
```
dsibilio, line 41: Are these used anywhere?

dsibilio (changes requested):

.github/actions/free-hosted-runner-disk-space/action.yml
```diff
@@ -1,18 +1,156 @@
 name: "Free GHA runner disk space"
 description: "Removes unnecessary files from the runner."
 inputs:
+  only-remove-files:
+    description: 'Only remove files and folders, do not manipulate the runner memory'
+    required: false
+    default: 'true'
```
dsibilio, line 7: IMO this needs a better name. Affirmative names rather than negative names are better and more intuitive. It is better to provide an optional flag to change the behaviour when its value is `true`, than providing an optional flag that you have to set to `false` to do the opposite of what the name says.

For example:
`merge-disk-volumes: true`

_(optional, default: false) to preserve backward compatibility_

version.txt
```diff
@@ -1 +1 @@
-v5.16.0
+v5.17.0
```
dsibilio, line 1: Conflict vs master needs to be handled.

.github/actions/free-hosted-runner-disk-space/action.yml
```diff
@@ -1,18 +1,156 @@
...
+    required: false
+    default: 'runner:runner'
+  pv-loop-path:
+    description: 'Absolute file path for the LVM image created on the root filesystem, the default is usually fine.'
+    required: false
+    default: '/pv.img'
+  tmp-pv-loop-path:
+    description: 'Absolute file path for the LVM image created on the temp filesystem, the default is usually fine. Must reside on /mnt'
+    required: false
+    default: '/mnt/tmp-pv.img'
```
dsibilio, line 42: IMO for all these properties the description can be enhanced to include that they're only applicable if `merge-disk-volumes` is `true`

.github/actions/free-hosted-runner-disk-space/action.yml
```diff
@@ -1,18 +1,156 @@
...
+        fi
+        if [[ -n "$to_remove" ]]; then
+          sudo rm -rf $to_remove
+        fi
+        echo "... done"
+      shell: bash
+
+    - name: Maximize build space
+      if: ${{ inputs.only-remove-files == 'false' }}
+      uses: easimon/maximize-build-space@v10
```
dsibilio, line 130: Please use [SHA Pinning](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions), example [here](https://github.com/Alfresco/alfresco-build-tools/blob/master/.github/actions/calculate-next-internal-version/action.yml#L22).

dsibilio (changes requested):

docs/README.md
```diff
@@ -525,20 +525,21 @@ referencing as value env vars defined early in the file (like Travis does).
 
 ### free-hosted-runner-disk-space
 
-Removes unnecessary folders and files from a GHA [hosted runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners). This action might be useful when we run jobs which require a lot of disk space.
+Removes unnecessary folders and files from a GHA [hosted runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners), and if that is not enough, allows to manipulate and utilize unused memory partitions. This action might be useful when we run jobs which require a lot of disk space.
```
dsibilio, line 528: We didn't add any documentation related to the new behaviour, at the least we should include the possibility to turn on the new behaviour via the newly implemented flag, passing through options to the underlying custom action and linking to the open source repository of that action so that people can look into it with ease.

dsibilio (changes requested):

docs/README.md
mikolajbrzezinski: Added a more in-depth description with an example below

docs/README.md
```diff
@@ -525,20 +525,35 @@ referencing as value env vars defined early in the file (like Travis does).
...
+This will allow to determine how much memory to allocate to filesystems, and where to mount the build volume, as shown in the example below.
+
+```yaml
+      - uses: Alfresco/alfresco-build-tools/.github/actions/free-hosted-runner-disk-space@ref
+        with:
+          merge-disk-volumes: true
+          root-reserve-mb: 12288
+          temp-reserve-mb: 100
+          swap-size-mb: 1024
+          build-mount-path: '/var/lib/docker/'
```
dsibilio, line 553:
```suggestion
          root-reserve-mb: 12288
          temp-reserve-mb: 100 # optional
          swap-size-mb: 1024 # optional
          build-mount-path: '/var/lib/docker/' # optional
```

docs/README.md
```diff
@@ -525,20 +525,35 @@ referencing as value env vars defined early in the file (like Travis does).
...
+```
+
+NOTE: When enabling [maximize-build-space](https://github.com/easimon/maximize-build-space) by setting `merge-disk-volumes` to `true`, this action should be used as the FIRST step, even before `actions/checkout`.
```
dsibilio, line 556:
```suggestion
> **NOTE:** when enabling [maximize-build-space](https://github.com/easimon/maximize-build-space) by setting `merge-disk-volumes` to `true`, this action should be used as the **FIRST** step, even before `actions/checkout`.
```

dsibilio (approved):
Left some minor commit suggestions, please review whether they make sense and in case apply them.

.github/actions/free-hosted-runner-disk-space/action.yml
```diff
@@ -1,18 +1,156 @@
...
+        echo
+        sudo swapon --show
+        echo
+
+        echo "Available storage:"
+        sudo df -h
+        echo
+
+    - name: Move /var/lib/docker/
+      if: ${{ inputs.merge-disk-volumes == 'true' }}
```
gionn, line 93:
```suggestion
      if: inputs.merge-disk-volumes == 'true'
```

.github/actions/free-hosted-runner-disk-space/action.yml
```diff
@@ -1,18 +1,156 @@
...
+          sudo rm -rf $AGENT_TOOLSDIRECTORY
+        fi
+        if [[ -n "$to_remove" ]]; then
+          sudo rm -rf $to_remove
+        fi
+        echo "... done"
+      shell: bash
+
+    - name: Maximize build space
+      if: ${{ inputs.merge-disk-volumes == 'true' }}
```
gionn, line 129:
```suggestion
      if: inputs.merge-disk-volumes == 'true'
```

.github/actions/free-hosted-runner-disk-space/action.yml
```diff
@@ -1,18 +1,156 @@
...
+        temp-reserve-mb: ${{ inputs.temp-reserve-mb }}
+        swap-size-mb: ${{ inputs.swap-size-mb }}
+        overprovision-lvm: ${{ inputs.overprovision-lvm }}
+        build-mount-path: ${{ inputs.build-mount-path }}
+        build-mount-path-ownership: ${{ inputs.build-mount-path-ownership }}
+        pv-loop-path: ${{ inputs.pv-loop-path }}
+        tmp-pv-loop-path: ${{ inputs.tmp-pv-loop-path }}
+
+    - name: Restore /var/lib/docker/
+      if: ${{ inputs.merge-disk-volumess == 'true' }}
```
gionn, line 142:
```suggestion
      if: inputs.merge-disk-volumess == 'true'
```

gionn (commented):
can you provide a link to a job with merge-disk-volumes enabled? I can't believe it could work without actually stop/start docker service before/after doing the mv of data

dsibilio:
**Note**: to fix pre-commit (extra whitespace at the end of the sentence): https://github.com/Alfresco/alfresco-build-tools/actions/runs/8187524258/job/22388181314?pr=524

.github/actions/free-hosted-runner-disk-space/action.yml
```diff
@@ -1,18 +1,156 @@
...
+      run: sudo sh -c "mv ${GITHUB_WORKSPACE}/docker/* /var/lib/docker"
```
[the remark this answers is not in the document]
mikolajbrzezinski: Unfortunately it doesn't work, already tested it before

gionn (approved):

PR #528, opened 2024-03-08
AAE-20957 Add actions to create deployments and update their status

atchertchian:
### Checklist

- Jira Reference (also in PR title): AAE-20957
  - [ ] Patch (bugfix)
  - [X] Minor (new feature)
- External PR link where changes has been tested: see mentions


gionn (commented):
much cool

.github/actions/github-deployment-status-update/action.yml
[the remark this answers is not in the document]
atchertchian: let me push another commit to improve this (as I agree, as visible in the README sample, that it's a lot of boilerplate for the basic use case)

.github/actions/github-deployment-status-update/action.yml
```diff
@@ -0,0 +1,42 @@
...
+  using: "composite"
+  steps:
+    - name: Update deployment status
+      id: update
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
+        DEPLOYMENT_ID: ${{ inputs.deployment-id }}
+        STATE: ${{ inputs.state }}
+        REPO_URL: ${{ github.server_url }}/${{ github.repository }}
+        FAILURE: ${{ contains(steps.*.outcome, 'failure') }}
```
atchertchian, line 25: this does not correctly allow to detect failure (job/step failure?) so not sure if this fallback can work after all...

.github/actions/github-deployment-status-update/action.yml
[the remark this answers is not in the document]
atchertchian: I could get simplify setting an initial state right after creation but other steps are easier to setup using condifitionals that can rely on `failure()` and similar helpers, so there's still some boilerplate left

docs/README.md
```diff
@@ -641,6 +643,46 @@ With proper concurrency logic in place, the latest run might have been cancelled
...
+        uses: Alfresco/alfresco-build-tools/.github/actions/github-deployment-create@ref
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          environment: my_gh_environment
+          state: in_progress
+
+      - name: Other Steps
+
+      - name: Update Deployment State to failure
+        if: always() && failure() && steps.create-deployment.outcome == 'success'
```
gionn, line 671: `always() && failure()` isn't equivalent to just `failure()`?

docs/README.md
```diff
@@ -641,6 +643,46 @@ With proper concurrency logic in place, the latest run might have been cancelled
...
+        uses: Alfresco/alfresco-build-tools/.github/actions/github-deployment-status-update@ref
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          deployment-id: ${{ steps.create-deployment.outputs.id }}
+          state: failure
+
+      - name: Update Deployment State to success
+        uses: Alfresco/alfresco-build-tools/.github/actions/github-deployment-status-update@ref
```
gionn, line 679:
```suggestion
        if: steps.create-deployment.outcome == 'success'
        uses: Alfresco/alfresco-build-tools/.github/actions/github-deployment-status-update@ref
```

otherwise it could cause an additional failure?

docs/README.md
atchertchian: sadly, no

docs/README.md
atchertchian: it is implied: if the step was not successful then a step without any condition will be skipped

docs/README.md
atchertchian: actually the doc at https://docs.github.com/en/actions/learn-github-actions/expressions#always recommend `if: ${{ !cancelled() }}`

`always() && failure()` is needed when you want a step in another job to be called even if previous job failed.

Anyway it's the readme, I can update

atchertchian: For the use case I documented you're right that it's not needed (and I tested) so I updated :+1:

.github/actions/github-deployment-create/action.yml
```diff
@@ -0,0 +1,47 @@
...
+        echo $CREATED
+
+        DEPLOYMENT_ID="$(echo $CREATED | jq -r ".id")"
+        echo "Created deployment $DEPLOYMENT_ID"
+
+        echo "id=$DEPLOYMENT_ID" >> $GITHUB_OUTPUT
+
+    - name: Update Deployment State
+      if: ${{ inputs.state != '' }}
+      uses: Alfresco/alfresco-build-tools/.github/actions/github-deployment-status-update@improvement/AAE-20957-deployment-status
```
atchertchian, line 43:
```suggestion
      uses: Alfresco/alfresco-build-tools/.github/actions/github-deployment-status-update@v5.18.0
```

@gionn should I apply the following for the release script to be happy, please?

gionn: yes with [skip ci] otherwise the build will fail

gionn (approved):

atchertchian:
will force-merge to accomodate for inner version ref management

PR #560, opened 2024-04-10
AAE-21850 Clean up skip PRs condition

tomgny:
### Checklist

  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)


mauriziovitale (approved):

atchertchian:
I'm starting to think that instead of adding so many exceptions to this list, since the "upstream" and "release" use cases are very specific to the single monorepo, maybe "whitelist-branches" should be used instead here, see doc at https://github.com/Alfresco/alfresco-build-tools?tab=readme-ov-file#enforce-pr-conventions
The only thing I'm not sure of is whether wildcards "upstream-*" and "release-*" would be handled correctly so that should be checked.

Overall I'd be in favor of keeping only "pr-*" in that line, that matches the generic dependabot use case (and was the original intent of this step).

WDYT please? cc @mauriziovitale @andrea-ligios

.github/actions/enforce-pr-conventions/action.yml
```diff
@@ -22,13 +22,13 @@ runs:
...
 
-    - name: Check is Dependabot PR or Propagation PR or Upstream PR
+    - name: Check is Dependabot PR or Propagation PR or Upstream PR or Release PR
       id: check
       env:
         PR_USER: ${{ github.event.pull_request.user.login }}
       shell: bash
       run: |
-          if [[ $PR_USER != 'dependabot[bot]' && $BRANCH_NAME != pr-* && $BRANCH_NAME != upstream-* ]]; then
+          if [[ $PR_USER != 'dependabot[bot]' && $BRANCH_NAME != pr-* && $BRANCH_NAME != upstream-* && $BRANCH_NAME != release-* ]]; then
```
gionn, line 31: there is a `whitelist-branches` input for avoiding hardcoding branches here, which can change depending on the target repo

gionn (changes requested):

.github/actions/enforce-pr-conventions/action.yml
mauriziovitale: as @atchertchian mentioned above most probably we need a wildcard as suffix
```  readarray -t branches <<< "${{ inputs.whitelist-branches }}"
        for branch in "${branches[@]}"; do
            if [[ $branch == $BRANCH_NAME ]]; then
            
  Or I am not sure if u can pass something like release-* as value

mauriziovitale (approved):

gionn (approved):

PR #586, opened 2024-04-24
do not try to release on dependabot pr close

gionn:
### Description

avoid failure like https://github.com/Alfresco/alfresco-build-tools/actions/runs/8812930250

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ jobs:
     release:
         name: Release
         runs-on: ubuntu-latest
-        if: github.actor != 'dependabot[bot]'
+        if: github.event.pull_request.merged == true
         steps:
           - uses: actions/checkout@1d96c772d19495a3b5c517cd2bc0cb401ea0529f # v4.1.3
             with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,7 @@ jobs:
     release:
         name: Release
         runs-on: ubuntu-latest
+        if: github.actor != 'dependabot[bot]'
         steps:
           - uses: actions/checkout@1d96c772d19495a3b5c517cd2bc0cb401ea0529f # v4.1.3
             with:


.github/workflows/release.yml
```diff
@@ -10,6 +10,7 @@ jobs:
     release:
         name: Release
         runs-on: ubuntu-latest
+        if: github.actor != 'dependabot[bot]'
```
atchertchian, line 13:
```suggestion
        if: github.event.sender.login != 'dependabot[bot]'
```

probably overkill here but handles more corner cases (re-run, or someone pushed commits to the PR, etc)

but I wonder if you want to avoid running on any PR which is closed (and not merged) anyway for this use case instead
so I'd suggest event:

```suggestion
        if: github.event.pull_request.merged == true
```

gionn: in general is not a problem to run the release workflow everytime, it won't have any side effects if the version is not incremented.

it's a problem when it run as dependabot because it hasn't access to secret which are required.

I never used `github.event.sender.login` and can't easily find some docs vs github.actor. I am pretty sure that if someone else interact with a dependabot PR then the actor is the user itself and not anymore dependabot. What's the different with event.sender?

atchertchian: it is the sender https://docs.github.com/en/webhooks/webhook-events-and-payloads

but I still don't follow: dependabot is the one closing the PR because there's a newest version avail, so you want to prevent the release workflow to be run

but you could prevent the release workflow to be run on PR closed (not merged, but both use cases are triggering the pr_closed event) for all use cases, not only dependabot: we don't want to run the release workflow unless the PR is merged instead of closed

atchertchian: the way I read the workflow right now, if I create a WIP branch where I bump the versions.txt file and I close this PR, a release will be triggered

overall I wonder if this release workflow should not be triggered on push on main branch instead (we can prevent it to be run again when commit is pushed), but just checking if "merged" bool is true should do it

dsibilio (approved):

atchertchian (approved):

PR #628, opened 2024-05-28
OPSEXP-2689 Add workflow_dispatch option for terraform

pmacius:
### Checklist

- Jira Reference (also in PR title): OPSEXP-2689
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)
- External PR link where changes has been tested: https://github.com/Alfresco/terraform-alfresco-nexus/pull/32


.github/workflows/terraform.yml
```diff
@@ -111,7 +115,7 @@ jobs:
 
       - name: Terraform plan
         uses: dflook/terraform-plan@0f4a5046f18698eee737a52d4cb11f56f2377a32 # v1.42.0
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request' || github.event.inputs.terrafom_operation == 'apply'
```
gionn, line 118: just `inputs.terrafom_operation` should be fine, see `terraform_root_path`

docs/README.md
```diff
@@ -1562,6 +1562,15 @@ on:
       - main
       - develop
   workflow_dispatch:
+    inputs:
+      terraform_option:
```
gionn, line 1566:
```suggestion
      terraform_operation:
```

and you need to propagate it in the workflow call as `with`

gionn (changes requested):

.github/workflows/terraform.yml
```diff
@@ -111,7 +115,7 @@ jobs:
 
       - name: Terraform plan
         uses: dflook/terraform-plan@0f4a5046f18698eee737a52d4cb11f56f2377a32 # v1.42.0
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request' || inputs.terraform_operation == 'apply'
```
gionn, line 118: I think it's not needed, as per docs it would just print plan on stdout outside of a pull_request event, but then the apply would be immediately executed after - without using that

https://github.com/dflook/terraform-plan?tab=readme-ov-file#terraform-plan-action

docs/README.md
```diff
@@ -1562,18 +1562,29 @@ on:
       - main
       - develop
   workflow_dispatch:
+    inputs:
+      terraform_operation:
+        description: 'Terraform option to perform apply or destroy operation.'
```
gionn, line 1567:
```suggestion
        description: 'Perform the requested operation on terraform'
```

gionn (changes requested):

.github/workflows/terraform.yml
pmacius: https://github.com/dflook/terraform-github-actions/tree/main/terraform-apply 
Without a plan apply will fail. Have to add auto_approve option based on event name.

.github/workflows/terraform.yml
```diff
@@ -124,9 +128,22 @@ jobs:
 
       - name: Terraform apply
         uses: dflook/terraform-apply@8a1bf89017460503cf302632cb15e8583afa7ac4 # v1.42.0
-        if: github.event_name == 'push'
+        if: github.event_name == 'push' || inputs.terraform_operation == 'apply'
         with:
           label: ${{ vars.RESOURCE_NAME }} ${{ inputs.terraform_root_path }}
+          path: ${{ inputs.terraform_root_path }}
+          auto_approve: ${{ github.event_name == 'workflow_dispatch' }}
```
gionn, line 135:
```suggestion
          auto_approve: ${{ inputs.terraform_operation == 'apply' }}
```

so you are sure that will be true only when requesting for apply

.github/workflows/terraform.yml
gionn: it could still work if the latest commit is a PR with a plan attached to it - but in that case you could simply re-run the merge workflow run - no need to trigger this new workflow.

anyway could be good to have it in case we know we want to run an apply.

gionn (changes requested):

gionn (approved):

gionn (approved):

PR #641, opened 2024-06-11
AAE-23095 Branch name length strategy

wojciech-piotrowiak:
### Checklist

  - [X] Patch (bugfix)
  - [ ] Minor (new feature)

### Description

Some branch names are too long, which causes tag name longer than maximum 128 characters


.github/actions/docker-build-image/action.yml
```diff
@@ -75,6 +75,7 @@ runs:
       run: |
         REPOSITORY=${DIR##*/}
         TAG="${TAG:-$(echo ${GITHUB_HEAD_REF:-$GITHUB_REF_NAME} | sed -e 's/[^-_.[:alnum:]]/_/g')}"
+        TAG="${TAG:0:128}"
```
atchertchian, line 78: Nitpick: if the tag originally comes from the action input, I don't think we should modify it, and let it fail if it's too long.
For the fallback tag (second part of expression 1 line above), we can cut it, but I wonder if we need to use the branch name on PRs, maybe relying on GITHUB_REF_NAME for all use cases is better ("<pr_number>/merge" would be the value on PRs)

wojciech-piotrowiak: thanks, set limitation only on generated value

atchertchian (approved):

PR #643, opened 2024-06-14
AAE-23198 Dependency Graph Submission on push in common build action

wojciech-piotrowiak:
### Checklist

  - [X] Patch (bugfix)
  - [ ] Minor (new feature)


.github/actions/maven-build-and-tag/action.yml
```diff
@@ -244,3 +244,10 @@ runs:
       with:
         path: ~/.m2/repository
         key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+
+    - name: Submit Dependency Graph
+      if: github.event_name == 'push'
+      uses: advanced-security/maven-dependency-submission-action@5d0f9011b55d6268922128af45275986303459c3 # v4.0.3
+      env:
+        MAVEN_USERNAME: ${{ inputs.maven-username }}
+        MAVEN_PASSWORD: ${{ inputs.maven-password }}
```
atchertchian, line 253: what about settings file?

please also rename the PR title to more meaningful

wojciech-piotrowiak: thanks, done
Did a test pr to check it and it works https://github.com/Alfresco/alfresco-identity-adapter-service/actions/runs/9515759065/job/26230694053?pr=925

atchertchian: looking at extra dependencies retrieved by this build, I wonder if it should be in the saved cache (so this step should be higher), wdyt?

also I'm not sure where the result is visible/uploaded: you gave a link to the PR but this is not supposed to be run on PRs, am I missing something?

wojciech-piotrowiak: I moved it even higher; that action may use dependencies removed on the end.

There is no direct clickable result, GH analyzes the output and creates security issues like this: https://github.com/Alfresco/hxp-process-services/security/dependabot/23

atchertchian (approved):

PR #655, opened 2024-06-18
AAE-23270 Print docker image built in GitHub summary

wojciech-piotrowiak:
### Checklist

  - [ ] Patch (bugfix)
  - [X] Minor (new feature)


.github/actions/docker-build-image/action.yml
```diff
@@ -209,3 +209,11 @@ runs:
         labels: ${{ steps.meta.outputs.labels }}
         push: true
         provenance: false
+
+    - name: Print tags
+      if: env.PUSH_IMAGE == 'true'
+      shell: bash
+      run: |
+        echo quay.io/alfresco/${{ env.IMAGE_NAME }}
```
atchertchian, line 217: why not print them in summary instead? (like on https://github.com/Alfresco/alfresco-build-tools/blob/65df92980d9506697dd4da49104d7d597bf27c2c/.github/actions/maven-build-and-tag/action.yml#L179)

wojciech-piotrowiak: Screenshot 2024-06-18 at 13 28 16
Thanks for a tip

.github/actions/docker-build-image/action.yml
```diff
@@ -209,3 +209,16 @@ runs:
...
+      if: env.PUSH_IMAGE == 'true'
+      shell: bash
+      run: |
+        {
+          echo "### :rocket: Docker image tags"
+          echo "| Arg                 | Value |"
+          echo "| ------------------- | ----- |"
+          echo "| Quay                | [quay.io/alfresco/${{ env.IMAGE_NAME }}](quay.io/alfresco/${{ env.IMAGE_NAME }}) |"
+          echo "| GHCR                | [ghcr.io/alfresco/${{ env.IMAGE_NAME }}](ghcr.io/alfresco/${{ env.IMAGE_NAME }}) |"
+          echo "| AWS ECR             | [${{ steps.login-ecr.outputs.registry }}/${{ env.IMAGE_NAME }}](${{ steps.login-ecr.outputs.registry }}/${{ env.IMAGE_NAME }}) |"
```
atchertchian, line 223: Looking at the test run at https://github.com/Alfresco/alfresco-identity-adapter-service/actions/runs/9564070539/attempts/1#summary-26364116425, the links do not actually work well (for instance the first one gives https://github.com/Alfresco/alfresco-identity-adapter-service/actions/runs/quay.io/alfresco/alfresco-identity-adapter-service:feature_AAE-23198-action-test): is giving the link actually useful? maybe an easy way to copy/paste it to a "docker pull" command would be better (?)

Also the table might look nice on adapter-service, where there's only one docker image built, but I think it should be much more compact to accommodate the process-services use case (34 images right now). Maybe we can have a try like this and see how it goes, though.

I would put the image name in the header, as suggested below (not sure if the image name in markdown title would look ok though)
If you think it's interesting to keep the links please fix them, maybe adding https:// in front is enough (and also "Arg" should be replaced by a meaningful column title)

```suggestion
          echo "### :rocket: Docker image built: \'${{ env.IMAGE_NAME }}\'"
```

wojciech-piotrowiak: My primary goal for tags was making it copyable, from my browser I can right click and copy (then paste it in rancher). Tried with different markdown options by probably links are the only way.

.github/actions/docker-build-image/action.yml
```diff
@@ -209,3 +209,16 @@ runs:
...
+      if: env.PUSH_IMAGE == 'true'
+      shell: bash
+      run: |
+        {
+          echo "### :rocket: Docker image built: \'${{ env.IMAGE_NAME }}\'"
+          echo "| Registry            | Image URL |"
+          echo "| ------------------- | ----- |"
+          echo "| Quay                | [quay.io/alfresco/${{ env.IMAGE_NAME }}](quay.io/alfresco/${{ env.IMAGE_NAME }}) |"
+          echo "| GHCR                | [ghcr.io/alfresco/${{ env.IMAGE_NAME }}](ghcr.io/alfresco/${{ env.IMAGE_NAME }}) |"
+          echo "| AWS ECR             | [${{ steps.login-ecr.outputs.registry }}/${{ env.IMAGE_NAME }}](${{ steps.login-ecr.outputs.registry }}/${{ env.IMAGE_NAME }}) |"
```
atchertchian, line 223: Ok I saw in your intermediate commits that you wanted to ease up copy/paste, and it's not working nicely with tables indeed, so here's another suggestion:

```suggestion
          echo "### :rocket: Docker image built:"
          echo "Quay:"
          echo "```\nquay.io/alfresco/${{ env.IMAGE_NAME }}\n```
          echo "GHCR:"
          echo "```\ghcr.io/alfresco/${{ env.IMAGE_NAME }}\n```
          echo "AWS ECR:"
          echo "```\n${{ steps.login-ecr.outputs.registry }}/${{ env.IMAGE_NAME }}\n```
```
If line returns are working as expected, this should give:

### :rocket: Docker image built
Quay:
```
quay.io/alfresco/alfresco-identity-adapter-service:feature_AAE-23198-action-test
```
GHCR:
```
ghcr.io/alfresco/alfresco-identity-adapter-service:feature_AAE-23198-action-test
```
AWS ECR:
```
645890235518.dkr.ecr.us-east-1.amazonaws.com/alfresco-identity-adapter-service:feature_AAE-23198-action-test
```

atchertchian (approved):

PR #656, opened 2024-06-18
HXOR-1490 Update helm-docs default version to 1.13.1

mteodori:
### Checklist

- Jira Reference (also in PR title):
- [README](https://github.com/Alfresco/alfresco-build-tools/blob/master/docs/README.md) updated after adding/changing behaviour of an action
- Proposed version increment for [release](https://github.com/Alfresco/alfresco-build-tools/blob/master/docs/README.md#release):
  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)
  - [ ] Major (breaking changes)
- External PR link where changes has been tested:

PR #657, opened 2024-06-19
HXOR-1490 Fix empty version input in setup-helm-docs

mteodori:
### Checklist

- Jira Reference (also in PR title): HXOR-1490
  - [X] Patch (bugfix)
  - [ ] Minor (new feature)
- External PR link where changes has been tested: test added in this PR

### Description
Fixes a regression introduced in #656 where an empty blank version input was generating an incorrect download link rather than using the default version of helm-docs


.github/actions/setup-helm-docs/action.yml
```diff
@@ -4,11 +4,13 @@ inputs:
   version:
     description: 'Version of helm-docs'
     required: false
-    default: 1.13.1
 runs:
   using: "composite"
   steps:
+    - shell: bash
+      run: |
+        echo "helm-docs-version=_${{ inputs.version }}_"
```
Mieszko96, line 12: Do we wanna keep this echo?

gionn, line 12:
```suggestion
```

mteodori: no, good catch, test leftovers, removing it now

.github/actions/setup-helm-docs/action.yml
```diff
@@ -4,11 +4,13 @@ inputs:
...
     - uses: Alfresco/alfresco-build-tools/.github/actions/setup-github-release-binary@v5.29.0
       with:
         repo: norwoodj/helm-docs
-        version: ${{ inputs.version }}
+        version: ${{ inputs.version != '' && inputs.version || '1.13.1' }}
```
gionn, line 16: maybe you can define it as env under `runs.env.DEFAULT_HELM_DOCS_VERSION` and reference it here as `env.DEFAULT_HELM_DOCS_VERSION`

.github/actions/setup-helm-docs/action.yml
```diff
@@ -4,11 +4,13 @@ inputs:
   version:
     description: 'Version of helm-docs'
     required: false
-    default: 1.13.1
```
alxgomz, line 7: Not sure I understand why using default input param is a problem

.github/tests/actions/test-setup-helm-docs/action.yml
```diff
@@ -0,0 +1,14 @@
...
+  helm-docs-version:
+    description: 'Version of helm-docs'
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - uses: ./.github/actions/setup-helm-docs
+      with:
+        version: ${{ inputs.helm-docs-version }}
```
alxgomz, line 14: Worth adding a small test to check version?

```suggestion
        version: ${{ inputs.helm-docs-version }}
    - name: Check Helm-docs version
       run: [ $(helm-docs -v | awk '{print $NF}') = ${{ inputs.helm-docs-version }} ]
```

.github/actions/setup-helm-docs/action.yml
gionn: https://github.com/Alfresco/alfresco-build-tools/blob/master/.updatecli/values/github_releases/helm_docs.yaml#L11 is also broken now

.github/actions/setup-helm-docs/action.yml
gionn: when you have a workflow which has an optional input invoking this actionoptional input, you would hope it will use the default value but instead will be evaluated to empty string

mteodori: it is not applied when called from another action/workflow where the version is an input and has an empty default, so either I repeat the same default on each or I do the check for empty

it did break here https://github.com/Alfresco/alfresco-process-infrastructure-deployment/actions/runs/9566428280/job/26406144776?pr=2214#step:5:120

gionn (approved):

PR #687, opened 2024-07-11
ACS-8392 Feature/GitHub changed files

tpage-alfresco:
### Checklist

- Jira Reference (also in PR title): ACS-8392
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)
- External PR link where changes has been tested: https://github.com/Alfresco/hxinsight-connector/pull/476

### Description

Previously we were using tj-actions/changed-files to determine the list of changed files in each PR, but during a Dependabot upgrade we found that we were not easily able to determine what had changed in the code. Consequently we decided it would be better to write our own action to avoid the potential supply chain risk.

More details are in the README, but the new action is able to process a `push` or `pull-request` event and output the list of changed files (optionally writing this to an env variable `GITHUB_MODIFIED_FILES`).


Pawel-608 (dismissed):
LGTM

krdabrowski (dismissed):
LGTM

mpichura (dismissed):

.github/actions/github-list-changes/action.yml
```diff
@@ -0,0 +1,26 @@
...
+  all_changed_files:
+    description: The list of changed files (one per line)
+    value: ${{ steps.list-changes.outputs.all-changed-files }}
+runs:
+  using: "composite"
+  steps:
+    - name: "Checkout"
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      with:
+        fetch-depth: 0
```
gionn, line 17: embedding checkout is usually not a wise choice because it would override any previous customization

if you need the `detch-depth: 0` for things to work, then add it  in the readme in the example snippet

gionn (commented):
> Previously we were using tj-actions/changed-files to determine the list of changed files in each PR, but during a Dependabot upgrade we found that we were not easily able to determine what had changed in the code. Consequently we decided it would be better to write our own action to avoid the potential supply chain risk.

I understand the point but I am against this approach but neither have an alternate solution to propose (IMHO using javascript is the main design failure of github actions).

So I am fine in maintaining this here but still, don't like it.

.github/actions/github-list-changes/action.yml
tpage-alfresco: Ah - ok. The `fetch-depth: 0` was exactly the reason to include this.

docs/README.md
```diff
@@ -718,6 +719,24 @@ Use this action when running a workflow which clone a private repository over ht
...
 ```
 
+### github-list-changes
+
+List the changes in a pull request (`pull-request` event) or that were pushed to a branch (`push` event).
+
+This action requires a checkout with fetch-depth option as follow:
+
+```yaml
+
```
gionn, line 729:
```suggestion
```

docs/README.md
```diff
@@ -718,6 +719,24 @@ Use this action when running a workflow which clone a private repository over ht
           pat: ${{ secrets.BOT_GITHUB_TOKEN }}
 ```
 
+### github-list-changes
+
+List the changes in a pull request (`pull-request` event) or that were pushed to a branch (`push` event).
+
+This action requires a checkout with fetch-depth option as follow:
```
gionn, line 726:
```suggestion
This action requires a checkout with `fetch-depth: 0` option as follow:
```

tpage-alfresco:
Retested with checkout moved to calling workflow: https://github.com/Alfresco/hxinsight-connector/pull/476

.github/actions/github-list-changes/github-list-changes.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
gionn, line 1:
```suggestion
#!/bin/bash -e
```

always add -e to make it resilient to random failures

krdabrowski (approved):
Still LGTM

PR #688, opened 2024-07-11
Ensure yq installed for dependabot pre-commit to work

gionn:
### Description

without `yq` installed you end up with an empty dependabot config after running it without any error.


tpage-alfresco (approved):
Thanks! I had to uninstall `yq` to test this, but worked like a charm :+1:

dsibilio (approved):

PR #691, opened 2024-07-18
ACSA-252: Switch SCA scan from maven to srcclr

mstrankowski:
### Checklist

  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)
- External PR link where changes has been tested: **needed**

### Description

Veracode is moving away from maven/gradle plugins and into their CLI tool "srcclr", we can also use the new tools ability to automatically create different projects in Veracode and link them to different application versions (despite same agent-based scan being set). So giving 1 additional input parameter on this action, project extension, should allow for neat linking of ACS_EXT_MASTER, ACS_EXT_MASTER_7_4, ACS_EXT_MASTER_7_3, ACS_EXT_MASTER_7_2 etc. with agent-based scan results from their branches without much manual work - just link the project to app in veracode and set the default branch for the project.

Still needs some work when we get information about setting maven options to skip test modules in scans. Waiting on veracode feedback on this.


.github/actions/veracode/action.yml
```diff
@@ -4,6 +4,11 @@ inputs:
   srcclr-api-token:
     description: "Agent API Token"
     required: true
+  srcclr-project-ext:
+    description: "Veracode project name extension"
+    required: false
+  # TODO: Add default options + custom options for Source Clear scan, need feedback from Veracode
+  # Below inputs are broken for now
```
dsibilio, line 11: I understand the 2 previous inputs are broken, but isn't there anything at all that we should implement so that each project - if needed - can pass their own custom options to the srcclr command?

.github/actions/veracode/source_clear.sh
```diff
@@ -4,10 +4,7 @@ echo "=========================== Starting SourceClear Script ==================
 PS4="\[\e[35m\]+ \[\e[m\]"
 set +e -v -x
 
-mvn -B -q clean install \
-    ${SRCCLR_INSTALL_DEFAULT_OPTIONS} ${SRCCLR_INSTALL_OPTIONS} \
-    com.srcclr:srcclr-maven-plugin:scan \
-    -Dcom.srcclr.apiToken=${SRCCLR_API_TOKEN} > scan.log
+srcclr scan > scan.log
```
dsibilio, line 7: Does this tool respect the https://github.com/Alfresco/alfresco-azure-connector/blob/master/srcclr.yml files that are scattered across our repositories? If not, this will be a significant issue.

.github/actions/veracode/source_clear.sh
```diff
@@ -4,10 +4,7 @@ echo "=========================== Starting SourceClear Script ==================
 PS4="\[\e[35m\]+ \[\e[m\]"
 set +e -v -x
 
-mvn -B -q clean install \
```
dsibilio, line 7: How does `SRCCLR_API_TOKEN` get injected now? Does the new `srcclr` command just expect it as an env var?

dsibilio (commented):
Personally I wouldn't merge this until:
- we test it on a repository that relies on a `srcclr.yml` file to exclude provided dependencies, and verify the findings make sense
- we test it on a repository that has test modules to skip _(e.g.: https://github.com/Alfresco/alfresco-transform-service/blob/b4e2855a8fb808104d40520061c310d74649e399/.github/workflows/ci.yml#L58 and https://github.com/Alfresco/hxinsight-connector)_
- we introduce more customization options if necessary _(allow arguments override / passthrough for the srcclr command via action inputs)_

version.txt
```diff
@@ -1 +1 @@
-v5.34.0
+v6.0.0
```
dsibilio, line 1: This is arguable and will depend on the solution, it might be also entirely possible that we manage to create something that is entirely backwards compatible workflow-wise, making this into a `5.35.0`! Needs more tests and info from Veracode related to features we use in some of our workflows **(e.g.: skipping some modules or using a specific Maven profile!)**

.github/actions/veracode/action.yml
```diff
@@ -14,10 +19,19 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: "Install srcclr CLI tool"
+      shell: bash
+      run: |
+        # Install srcclr CLI tool (Veracode) as per https://docs.veracode.com/r/t_sc_cli_agent
+        sudo apt-get update
+        sudo apt-get install srcclr
```
gionn, line 27:
```
curl -sSL 'https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xdf7dd7a50b746dd4' | sudo gpg --dearmor -o /usr/share/keyrings/veracode-sca-archive.gpg
echo 'deb [signed-by=/usr/share/keyrings/veracode-sca-archive.gpg] https://download.sourceclear.com/ubuntu stable/' | sudo tee /etc/apt/sources.list.d/veracode-sca.list
sudo apt-get update
sudo apt-get install srcclr   
```

.github/actions/veracode/source_clear.sh
gionn: https://docs.veracode.com/r/Veracode_SCA_Agent_Environment_Variables

.github/actions/veracode/source_clear.sh
gionn: https://docs.veracode.com/r/Veracode_SCA_Agent_Commands#scan

should be in the same folder of srccclr

mstrankowski: Yes it does, in fact it will require the "srcclr-install-options" custom values to be placed in that same file to work, so will require a slight rework for a few projects (HxInsight, ESC, ATS, maybe Audit component "coming soon") that were using this option until now.

.github/actions/veracode/action.yml
mstrankowski: Will do

.github/actions/veracode/source_clear.sh
mstrankowski: Yep, it uses env variables or stuff it can retrieve from github repository on its own, I'm only overriding the URI there to use the automatic project creation mechanism.

.github/actions/veracode/action.yml
```diff
@@ -4,20 +4,27 @@ inputs:
   srcclr-api-token:
     description: "Agent API Token"
     required: true
-  srcclr-install-default-options:
-    description: Default options for Source Clear scan
-    required: false
-    default: '-DskipTests -Dmaven.javadoc.skip=true'
-  srcclr-install-options:
-    description: Additional options for Source Clear scan
```
mstrankowski, line 12: "srcclr-install-default-options" disappears, I've tested the srcclr CLI tool locally to make sure it doesn't run the tests when executing a scan, seems that it doesn't so no need for this default option.

"srcclr-install-options" Will require (described in readme) a small rework to use 
https://docs.veracode.com/r/Java_Scan_Directives#custom_maven_command
or
https://docs.veracode.com/r/Java_Scan_Directives#custom_maven_exec
To replace current behaviour for HXInsight, ESC, ATS and potentially for future Audit storage component.

.github/actions/veracode/action.yml
```diff
@@ -4,20 +4,27 @@ inputs:
...
+        # Install srcclr CLI tool (Veracode) as per https://docs.veracode.com/r/t_sc_cli_agent
+        curl -sSL 'https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xdf7dd7a50b746dd4' | sudo gpg --dearmor -o /usr/share/keyrings/veracode-sca-archive.gpg
+        echo 'deb [signed-by=/usr/share/keyrings/veracode-sca-archive.gpg] https://download.sourceclear.com/ubuntu stable/' | sudo tee /etc/apt/sources.list.d/veracode-sca.list
+        sudo apt-get update
+        sudo apt-get install srcclr
+    - name: "Add Veracode project name extension"
+      shell: bash
+      if: srcclr-project-ext != ''
+      run: |
+        echo "SRCCLR_SCM_URI=${{ github.server_url }}/${{ github.repository }}/${{ inputs.srcclr-project-ext }}" >> $GITHUB_ENV
```
mstrankowski, line 25: Adding such a suffix lets us use action for 
alfresco_enterprise_repo for branches 23.N, 7.4.N, 7.3.N and 7.2.N.
Without it we'd get results from all of them in the same project on Veracode side.
Adding "7.4.N" as srcclr-project-ext, will create 
alfresco_enterprise_repo_7.4.N project on veracode side, which can be conviniently linked to application version in Veracode. Just set a default branch for this project - release/7.4.N and link it to the ACS_EXT_MASTER_7_4 app.

version.txt
mstrankowski: Yeah, those need to be placed in srcclr.yml file in the respective repository with this approach, can't really be in command execution, but you could theoretically hack it to use the older input and modify the srcclr.yml file in this action before the scan is ran with sed or something? I'm not 100% sure that would work though.

.github/actions/veracode/action.yml
mstrankowski: There might be - modify the projects srcclr.yml file with the action to move the value from input of action into that location, a bit cumbersome and complicated imo, don't know if it's worth it necessarily.

.github/actions/veracode/source_clear.sh
gionn, line 7: maybe now this scan.log trickery is not needed anymore?

I guess it was there because maven output was a lot more verbose before

mstrankowski: Good point, but I'd actually prefer going the other way. 
I just added the "--loud" option to make the logging a lot more verbose and left the redirect, will be a tad easier to start investigation if something fails.

version.txt
mstrankowski: I've tested with alfresco-transform-service, on master branch

I've added a line in srcclr.yml of that project
`custom_maven_command: clean install -DskipTests -pl "!test-end-to-end,!test-load"`
and I've used command: 
`srcclr scan --loud --allow-dirty`

Result:
Full Report Details                            https://sca.analysiscenter.veracode.com/teams/QmmUV9P/scans/70631289

Also no mention of testing modules in srcclr execution logs.

--loud option is to print full logs, 
--allow-dirty allows for running scans on non-commited changes - in case we decide to modify this file with action while keeping it backwards compatible. Theoretically we could also push the change, but that might complicate it.

dsibilio: I'm ok with doing it the clean way _(by editing the `srcclr.yml` file in reach repository)_ and having this as a breaking change.

.github/actions/veracode/source_clear.sh
```diff
@@ -4,10 +4,7 @@ echo "=========================== Starting SourceClear Script ==================
...
-    ${SRCCLR_INSTALL_DEFAULT_OPTIONS} ${SRCCLR_INSTALL_OPTIONS} \
-    com.srcclr:srcclr-maven-plugin:scan \
-    -Dcom.srcclr.apiToken=${SRCCLR_API_TOKEN} > scan.log
+srcclr scan --loud > scan.log
```
dsibilio, line 7: Did we test this on a project that has actual SCA findings, to verify that the `grep ...` instruction few lines before still works as intended and fetches the Veracode Agent Based Scan URL?

docs/README.md
```diff
@@ -1537,12 +1537,15 @@ Validates Maven dependency graph versions to ensure all target includes artifact
...
+To add custom maven execution instructions, for skipping test modules etc., use [custom_maven_command](https://docs.veracode.com/r/Java_Scan_Directives#custom_maven_command)
+scan directive in srcclr.yml file of the scanned repository.
+
 ```yaml
       - uses: Alfresco/alfresco-build-tools/.github/actions/veracode@ref
         #continue-on-error: true # uncomment this line to prevent the Veracode scan step from failing the whole build
         with:
           srcclr-api-token: ${{ secrets.SRCCLR_API_TOKEN }}
-          srcclr-install-options: '-DskipTestModules' # optional, additional maven options
+          srcclr-project-ext: '' # optional, sets Veracode project extension
```
dsibilio, line 1548: I don't think from here it'll be clear what this project extension is for, IMO we need more documentation

.github/actions/veracode/action.yml
```diff
@@ -4,20 +4,27 @@ inputs:
...
     description: "Agent API Token"
     required: true
-  srcclr-install-default-options:
-    description: Default options for Source Clear scan
-    required: false
-    default: '-DskipTests -Dmaven.javadoc.skip=true'
-  srcclr-install-options:
-    description: Additional options for Source Clear scan
+  srcclr-project-ext:
+    description: "Veracode project name extension"
```
dsibilio, line 8: I don't think from here it'll be clear what this project extension is for, IMO we need more documentation

.github/actions/veracode/source_clear.sh
mstrankowski: Changed my mind after Domenico's comments, I'll just do your approach. All that the --loud option does is show maven logs, where it fails once, because it tries to get stuff without actually building, then when it sees that it failed, it builds the project and uploads that. Having these fails might actually be more confusing than not having logs for maven at all.

.github/actions/veracode/source_clear.sh
mstrankowski: It does, but the "-e 'Failed'" gets stuff from maven build because srcclr tries a cheeky way of getting a build in a low-cost manner first, before it does a proper one and the first attempt fails.

I've went back on my decision of adding "--loud" there and I followed Giovanni's recommendation to just remove the script and just call the scan in actions step itself.

Truth be told the only thing that this --loud provides over normal call, are maven logs, and unless that fails I don't think we should bother too much. It can always be tested locally if need arises.

mstrankowski: Also, we get a nicely formatted report without even having to use grep.

docs/README.md
mstrankowski: I've tried to be more descriptive and lined out sample of the usage in readme file

.github/actions/veracode/action.yml
mstrankowski: I've tried to be more descriptive and lined out sample of the usage in readme file

gionn (approved):

.github/actions/veracode/source_clear.sh
dsibilio: I'd like to see 1 example of such logs to make sure we get at least the same level of information we had before (the link to hit the Veracode Web UI) 👍

dsibilio: Actually, given that the logs contain sensitive information (e.g.: the vulnerabilities we're affected by) and I'm pretty confident any registered user with read access to a GitHub Repository can also read the logs of a GH Action Run, we should retain the logic to redirect the output somewhere else and only grep for the "Full Report Details" in case of failure! cc: @gionn

.github/actions/veracode/action.yml
mstrankowski, line 1: Tested on 
https://github.com/Alfresco/alfresco-enterprise-repo/pull/2125/checks

.github/actions/veracode/action.yml
```diff
@@ -4,20 +4,35 @@ inputs:
...
+        echo "SRCCLR_SCM_URI=$SRCCLR_SCM_URI" >> $GITHUB_ENV
     - name: "Run scanning"
       run: ${{ github.action_path }}/source_clear.sh
       shell: bash
       env:
         SRCCLR_API_TOKEN: ${{ inputs.srcclr-api-token }}
-        SRCCLR_INSTALL_DEFAULT_OPTIONS: ${{ inputs.srcclr-install-default-options }}
-        SRCCLR_INSTALL_OPTIONS: ${{ inputs.srcclr-install-options }}
+        SRCCLR_SCM_REF: ${{ github.event_name == 'pull_request' && github.ref || github.ref_name }}
+        SRCCLR_SCM_REF_TYPE: ${{ github.event_name == 'pull_request' && 'tag' || 'branch' }}
```
dsibilio, line 37: There are ways to fetch the original branch name from a PR event if needed, using `github.head_ref` instead of `github.ref`!

Also, I'm not sure I understand the logic related to PR = tag and rest = branch?

mstrankowski: I know it is possible, but I don't think that's what we want. 

A PR is a combination of the current master + changes proposed, a branch has the old master base and new changes proposed. I don't want to mix the 2.

If I target a branch instead of a tag, I will be showing potentially appeared/removed issues from new base as if it was on the old base. Gets a bit confusing.
Might as well use the tag then, it works fine, shows vulnerabilities properly in Veracode scans and doesn't confuse the base branch results with itself, especially given that current PR SCA scans just throw 

`Error:  Failed to execute goal com.srcclr:srcclr-maven-plugin:3.1.25:scan (default-cli) on project alfresco-community-repo: Could not perform scan because an apiToken was not provided. Please provide it either through the "apiToken" Maven parameter or the "SRCCLR_API_TOKEN" environment variable -> [Help 1]`

dsibilio: Alright, sounds like you planned this appropriately then. LGTM!

dsibilio (approved):
Great work! LGTM.

IMO it would be great to have an issue or a Confluence page where we document the trickiest parts of migrating from build-tools 5.x to 6.x due to this change.

E.g.: how to deal with test modules, custom Maven commands, and whatever else may not be immediately apparent from this PR

gionn (approved):

PR #702, opened 2024-07-31
OPSEXP-2758 Drop workaround for setuptools incident after upstream fix

gionn:
### Checklist

- Jira Reference (also in PR title): OPSEXP-2758

### Description

A new version of setuptools has been released which doesn't wreak the havoc for libraries still using the now deprecated feature.

https://github.com/pypa/setuptools/blob/main/NEWS.rst#v7210


dsibilio (approved):
Great stuff

PR #725, opened 2024-08-28
AAE-24324 Adapt Management Of Nexus Staging Repositories to Nexus 3

revatilimaye:
### Checklist

- Jira Reference (also in PR title): https://hyland.atlassian.net/browse/AAE-24324
  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)

### Description
Issue Link: https://hyland.atlassian.net/browse/AAE-24324


docs/README.md
```diff
@@ -1042,6 +1045,49 @@ The resulting staging repository will be available in the output named `staging-
...
 ```
 
+### nexus-create-tag
+
+Creates a tag on Nexus with a specific name. The resulting tag will be available in the output named `tag-name`.
+
+```yaml
+      - uses: Alfresco/alfresco-build-tools/.github/actions/nexus-create-tag@ref
+        with:
+          tag-name:  ${{ needs.load-release-info.outputs.version }}
```
gionn, line 1055: provide an example value, this has no meaning in this context

docs/README.md
```diff
@@ -1042,6 +1045,49 @@ The resulting staging repository will be available in the output named `staging-
...
+```
+
+### nexus-associate-tag
+
+Associates tag to artifacts on Nexus. The artifacts are filtered by provided maven group-id and version and then associated with the provided tag.
+
+```yaml
+      - uses: Alfresco/alfresco-build-tools/.github/actions/nexus-associate-tags@ref
+        with:
+          tag-name: ${{needs.load-release-info.outputs.version }}
```
gionn, line 1067: provide an example value, maybe the same used in nexus-create-tag example if it's the same

.github/actions/nexus-associate-tag/action.yml
```diff
@@ -0,0 +1,42 @@
...
+        NEXUS_PASSWORD: ${{ inputs.nexus-password }}
+        NEXUS_URL: ${{ inputs.nexus-url }}
+        GROUP-ID: ${{ inputs.group }}
+        VERSION: ${{ inputs.version }}
+      run: |
+          echo "Associating tag $TAG_NAME with component $GROUP:$VERSION"
+            curl -u "${NEXUS_USERNAME}":"${NEXUS_PASSWORD}" \
+                -X POST "${NEXUS_URL}/service/rest/v1/tags/associate/${TAG_NAME}&group=${GROUP-ID}&version=${VERSION}" \
+                -H "Content-Type:application/json" \
+                -d '{}'
```
gionn, line 42: wrong indent

also make sure to add `-sSf` to every curl command otherwise the step is not gonna fail if the response is not 2xx

docs/README.md
```diff
@@ -1042,6 +1045,49 @@ The resulting staging repository will be available in the output named `staging-
           nexus-password: "${{ secrets.NEXUS_PASSWORD }}"
 ```
 
+### nexus-create-tag
+
+Creates a tag on Nexus with a specific name. The resulting tag will be available in the output named `tag-name`.
```
gionn, line 1050: I don't see any output being set, probably because it doesn't make sense to have one if it's the same value provided as input

.github/actions/nexus-associate-tag/action.yml
```diff
@@ -0,0 +1,42 @@
...
+  nexus-username:
+    description: Nexus user name
+    required: true
+  nexus-password:
+    description: Nexus password
+    required: true
+  nexus-url:
+    description: Base URL to the Nexus server
+    required: false
+    default: "https://artifacts.alfresco.com/nexus"
```
gionn, line 16: better avoid hardcoding urls in input, we can add a global variable at the org level and use that in the downstream repositories

.github/actions/nexus-associate-tag/action.yml
```diff
@@ -0,0 +1,42 @@
...
+  group-id:
```
erdemedeiros, line 17: I think you can simplify it to `group`

.github/actions/nexus-associate-tag/action.yml
```diff
@@ -0,0 +1,42 @@
...
+  steps:
+    - name: associate-tag
+      id: associate-tag
+      shell: bash
+      env:
+        TAG_NAME: ${{ inputs.tag-name }}
+        NEXUS_USERNAME: ${{ inputs.nexus-username }}
+        NEXUS_PASSWORD: ${{ inputs.nexus-password }}
+        NEXUS_URL: ${{ inputs.nexus-url }}
+        GROUP-ID: ${{ inputs.group }}
```
erdemedeiros, line 35: AFAIK `-` are not allowed in variable names and I think yo can simplify it to `GROUP`

.github/actions/nexus-move-artifacts/action.yml
```diff
@@ -0,0 +1,38 @@
+name: Move artifacts to destination repository
+description: Move artifacts to destination repository in Nexus
+inputs:
+  tag-name:
+    description: The name of the tag
+    required: true
+  nexus-username:
+    description: Nexus user name
```
erdemedeiros, line 8:
```suggestion
    description: Nexus username
```

.github/actions/nexus-move-artifacts/action.yml
```diff
@@ -0,0 +1,38 @@
...
+        NEXUS_USERNAME: ${{ inputs.nexus-username }}
+        NEXUS_PASSWORD: ${{ inputs.nexus-password }}
+        NEXUS_URL: ${{ inputs.nexus-url }}
+        DESTINATION_REPOSITORY: ${{ inputs.destination-repository }}
+      run: |
+        echo "Move artifacts with tag $TAG_NAME to staging repository $STAGING_REPOSITORY"
+        curl -u "${NEXUS_USERNAME}:${NEXUS_PASSWORD}" \
+          -X POST "${NEXUS_URL}/service/rest/v1/tags/move/{DESTINATION_REPOSITORY}?tag=${TAG_NAME}" \
+          -H 'accept: application/json' \
+          -d '{}'
```
erdemedeiros, line 38: I think you also need o specify a `source-repository` (`repository` parameter in the API)

docs/README.md
```diff
@@ -70,6 +70,9 @@ Here follows the list of GitHub Actions topics available in the current document
     - [maven-release](#maven-release)
     - [maven-update-pom-version](#maven-update-pom-version)
     - [nexus-create-staging](#nexus-create-staging)
```
erdemedeiros, line 72: I think the actions `nexus-create-staging`, `nexus-close-staging` and `nexus-release-staging` can be remove because they only work on Nexus 2 and we only use Nexus 3 now.

.github/actions/nexus-associate-tag/action.yml
atchertchian, line 42: Can't the group and version be passed here instead? (allows proper encoding of query parameters)

.github/actions/nexus-create-tag/action.yml
```diff
@@ -0,0 +1,39 @@
...
+        NEXUS_USERNAME: ${{ inputs.nexus-username }}
+        NEXUS_PASSWORD: ${{ inputs.nexus-password }}
+        NEXUS_URL: ${{ inputs.nexus-url }}
+      run: |
+          echo "Creating tag $TAG_NAME in Nexus"
+          curl -u "${NEXUS_USERNAME}:${NEXUS_PASSWORD}" \
+            -X POST --header 'Content-Type: application/json' \
+            "${NEXUS_URL}/service/rest/v1/tags" \
+            -d "{\"name\":\"$TAG_NAME\"}"
+          echo "tag-name-created=$TAG_NAME" >> $GITHUB_OUTPUT
```
erdemedeiros, line 39: I think you can simplify the output name to `tag`

version.txt
```diff
@@ -1 +1 @@
-v6.1.0
+v6.2.0
```
erdemedeiros, line 1: If the previous actions are deleted, it's better to bump the major, i.e. `v7.0.0`

erdemedeiros (changes requested):

.github/actions/nexus-associate-tag/action.yml
erdemedeiros: Good catch!

erdemedeiros: Sorry, @atchertchian, I don't get what you mean here. Group and version are already passed as inputs and mapped to env vars

.github/actions/nexus-create-tag/action.yml
erdemedeiros: Update: @gionn mentioned, it probably not needed provide an output in this case because we already know the value which is the same passed as input.

docs/README.md
erdemedeiros: I agree, I don't think an out put is needed here

docs/README.md
```diff
@@ -1042,6 +1045,49 @@ The resulting staging repository will be available in the output named `staging-
...
+Associates tag to artifacts on Nexus. The artifacts are filtered by provided maven group-id and version and then associated with the provided tag.
+
+```yaml
+      - uses: Alfresco/alfresco-build-tools/.github/actions/nexus-associate-tags@ref
+        with:
+          tag-name: ${{needs.load-release-info.outputs.version }}
+          nexus-username: ${{ secrets.NEXUS_USERNAME }}
+          nexus-password: ${{ secrets.NEXUS_PASSWORD }}
+          group-id: maven-group-id
+          version: ${{ steps.load-descriptor.outputs.version }}
```
erdemedeiros, line 1071:
```suggestion
          tag-name: 1.0.0
          nexus-username: ${{ secrets.NEXUS_USERNAME }}
          nexus-password: ${{ secrets.NEXUS_PASSWORD }}
          group-id: com.company
          version: 1.0.0
```

.github/actions/nexus-associate-tag/action.yml
atchertchian: It's a detail: query params can be passed separately so that they can be properly encoded, but that does not seem valid for POST after a quick try, and seems overkill given potential values for params, so nevermind

gionn (commented):
just a few rewords in the readme, plus specifying the version of nexus which is important for the compatibility of those new actions

gionn (commented):
what's the rationale behind the choice of dropping the tag feature?

.github/actions/nexus-count-artifacts/action.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      run: |
+        echo "Counting artifacts in repository ${REPOSITORY} in Nexus"
+        search_response=$(curl -sSf -u "${NEXUS_USERNAME}:${NEXUS_PASSWORD}" \
+          -X GET "${NEXUS_URL}/service/rest/v1/search?repository=${REPOSITORY}&group=${GROUP}&version=${VERSION}" \
+          -H 'Content-Type: application/json' \
+          -H 'accept: application/json' \
+          -d '{}')
+        artifact_count=$(echo "$search_response" | yq '.items | length')
+        echo "Artifact Count: $artifact_count"
```
erdemedeiros, line 49: Actually, depending on the number of the artifacts the results can be [paged](https://help.sonatype.com/en/pagination.html). This is not problem for our use case where we just need to know if there are elements to move or not, but I was wondering if we should add a marker in case where the `continuationToken` is not `null` (meaning that there are more elements than the ones returned). For instance, return `100` if  `continuationToken` is `null` and return `100+` if it's not `null`. 

Alternatively, we could embed the call directly in the action `nexus-move-artifact` instead of having a dedicated one for counting.

What do you think @revatilimaye @atchertchian @gionn ?

.github/actions/nexus-move-artifacts/action.yml
```diff
@@ -0,0 +1,47 @@
+name: Move artifacts to destination repository
+description: Move artifacts to destination repository in Nexus
```
erdemedeiros, line 2:
```suggestion
description: Move artifacts to destination repository in Nexus 3
```

.github/actions/nexus-count-artifacts/action.yml
```diff
@@ -0,0 +1,50 @@
+name: Count artifacts in Nexus Repository
+description: Count artifacts in Nexus Repository
```
erdemedeiros, line 2:
```suggestion
description: Count artifacts in Nexus 3 Repository. In the case where too many elements matches with given criteria, the count will represent only the number of elements in the first page
```

erdemedeiros (commented):
I added a open question and two small suggestions. LGTM, otherwise

.github/actions/nexus-count-artifacts/action.yml
atchertchian: I have a preference for this action being embedded in the "move" one

revatilimaye: Have embedded the count-artifacts action in move-artifacts

erdemedeiros:
>what's the rationale behind the choice of dropping the tag feature?
@gionn, while reading the [migration guidelines](https://help.sonatype.com/en/upgrading-staging.html#path-from-nexus-repository-2-to-3) our initial understanding was that it was necessary to create a tag in Nexus and associate artifacts to this tag to be able to move artifacts from one repository to another. However, after experimenting with the actual API, we noticed that using `groupId` and `version` was enough for your use case, so no point on creating and maintaining two additional endpoints if they are not needed.

erdemedeiros (approved):
LGTM!

PR #761, opened 2024-10-02
OPSEXP-2801: refactor and bugfix of dbp-charts/verify-compose action

alxgomz:
### Checklist

- Jira Reference (also in PR title): OPSEXP-2801
  - [X] Patch (bugfix)
  - [ ] Minor (new feature)
- External PR link where changes has been tested: https://github.com/Alfresco/alfresco-dockerfiles-bakery/actions/runs/11147343637/job/30984260902?pr=53

### Description

* Refactor to minimize dup & get logs on compose errors
* Fix compose failing when no `docker-compose.yml` is present
* Do not wait for Solr is ACS is not configured to use it as a search backend


.github/actions/dbp-charts/verify-compose/docker_compose.sh
```diff
@@ -5,6 +5,26 @@ COMPOSE_PATH=$(dirname $COMPOSE_FILE_PATH)
 COMPOSE_BIN="docker compose"
 alf_port=8080
 
+containers_dump_logs_on_error() {
```
pmacius, line 8: I would use different name - is it only dumping logs when on error?

alxgomz: indeed not, it's only being called on errors in the script XD

pmacius (approved):

PR #807, opened 2024-11-05
AAE-25854 New docker build action

wojciech-piotrowiak:
### Checklist

  - [ ] Patch (bugfix)
  - [x] Minor (new feature)

### Description

We faced issue where one token should be used to pull docker images during the build and other one with write permission should be used to push. 
Similar approach is used in hxpr: https://github.com/HylandSoftware/hxpr/blob/master/.github/workflows/continuous_integration.yaml#L197
https://github.com/HylandSoftware/hxpr/blob/master/.github/workflows/continuous_integration.yaml#L217

New action does not require keepin backward compability and also forces users to provide ghcr username which is crucial as we want to avoid using ${{ github.actor }}


gionn (changes requested):
1. why a new action which looks like a fork of `docker-build-image`?
2. this action is so tied to your specific builds that I am becoming concerned that should not be published anymore here under build-tools but on a private repository

wojciech-piotrowiak:
@gionn 
1. it looks like a fork because we need a lot of logic inside however refactoring current action to have backward compatibility will be painful
2. This new action will be used in at least 3 different repos so I need a common place anyway

gionn:
this has been moved to the new internal-build-tools

PR #860, opened 2025-01-08
AAE-29961 Fix link to reportportal launch when using ubuntu 24

atchertchian:
Just use python to urlencode instead of curl

### Checklist

  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)


dsibilio (approved):
pain

PR #861, opened 2025-01-09
AAE-28943 Include sonar scanner in build action

mavotto:
### Checklist

  - [ ] Patch (bugfix)
  - [X] Minor (new feature)
https://github.com/Alfresco/hxp-studio-services/pull/3306


version.txt
```diff
@@ -1 +1 @@
-v8.8.1
+v8.9.1
```
pmacius, line 1:
```suggestion
v8.9.0
```

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -240,13 +257,19 @@ runs:
         echo "testcontainers.reuse.enable=true" > ~/.testcontainers.properties
         echo "TESTCONTAINERS_RYUK_DISABLED=true" >> $GITHUB_ENV
 
+    - name: Set SONAR_SCANNER_OPTS
+      if: ${{ inputs.sonar-token != '' && github.actor != 'dependabot[bot]' }}
```
pmacius, line 261: Im not sure about setting this here on this level. How about adding simple true false switch from inputs and check if we want to run this step in upstream workflow. @gionn

version.txt
mavotto: I followed this information:

`New versions should follow [Semantic versioning](https://semver.org/), so:

A bump in the third number will be required if you are bug fixing an existing action.
A bump in the second number will be required if you introduced a new action or improved an existing action, ensuring backward compatibility.
A bump in the first number will be required if there are major changes in the repository layout, or if users are required to change their workflow config when upgrading to the new version of an existing action.`

Since this is not a bug fix but an improvement of an existing action, I bumped the second number. I'm asking just to understand. Thanks.

pmacius: Check https://semver.org/#spec-item-7

.github/actions/maven-build-and-tag/action.yml
mavotto:
```suggestion
            if: ${{ inputs.enable-sonar-scan && github.actor != 'dependabot[bot]' }}
```
like this?

mavotto: I need the check on the dependabot because I don't want to execute a sonar scan in the dependabot pr

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -240,13 +257,19 @@ runs:
...
+      shell: bash
+      run: echo "SONAR_SCANNER_OPTS=sonar:sonar -Dsonar.host.url=${{ inputs.sonar-host-url }} -Dsonar.organization=${{ inputs.sonar-organization }} -Dsonar.projectKey=${{ inputs.project-key }} -Dsonar.coverage.jacoco.xmlReportPaths=${{ inputs.aggregate-report-path }}" >> $GITHUB_ENV
+
     - name: Build and Test with Maven (and maybe Deploy)
       shell: bash
-      run: mvn ${{ steps.define_maven_command.outputs.command }} ${{ env.MAVEN_CLI_OPTS}} ${{ inputs.extra-maven-opts }}
+      run: mvn ${{ steps.define_maven_command.outputs.command }} ${{ env.MAVEN_CLI_OPTS}} ${{ inputs.extra-maven-opts }} ${{ env.SONAR_SCANNER_OPTS }}
```
gionn, line 267:
```suggestion
      run: mvn ${{ steps.define_maven_command.outputs.command }} ${{ env.MAVEN_CLI_OPTS }} ${{ inputs.extra-maven-opts }} ${{ env.SONAR_SCANNER_OPTS }}
```

.github/actions/maven-build-and-tag/action.yml
gionn: no need to explicitly checks for github.actor, dependabot would not have access to sonar-token anyway.

I would just have:

```suggestion
      if: ${{ inputs.sonar-token != '' }}
```

boolean help in being explicit but in this case we have a secret and really make it straight-forward to use just that

.github/actions/maven-build-and-tag/action.yml
gionn: make sure to test on a real workflow what happen when SONAR_SCANNER_OPTS is not defined because I am not sure it would just evaluated it as empty string here

mavotto: I tested it here https://productionresultssa16.blob.core.windows.net/actions-results/8b076646-22b3-4bf1-8ec3-f8d081f942c0/workflow-job-run-ca395085-040a-526b-2ce8-bdc85f692774/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-01-09T15%3A11%3A29Z&sig=jryQyba8ULA2td7L6pcTyyiFS4s%2BYTr2PGvyiDAMLaY%3D&ske=2025-01-10T00%3A35%3A22Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-01-09T12%3A35%3A22Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-11-04&sp=r&spr=https&sr=b&st=2025-01-09T15%3A01%3A24Z&sv=2024-11-04

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -240,13 +257,19 @@ runs:
         echo "testcontainers.reuse.enable=true" > ~/.testcontainers.properties
         echo "TESTCONTAINERS_RYUK_DISABLED=true" >> $GITHUB_ENV
 
+    - name: Set SONAR_SCANNER_OPTS
+      if: ${{ inputs.sonar-token != '' }}
```
gionn, line 261:
```suggestion
      if: inputs.sonar-token != ''
```

gionn (approved):

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -86,6 +86,23 @@ inputs:
     description: The label name for skipping tests
     required: false
     default: 'skip-tests'
+  sonar-token:
+    description: 'SonarCloud token'
+    required: false
+  aggregate-report-path:
+    description: 'Path of the aggregated JaCoCo report'
+    required: false
+  project-key:
```
atchertchian, line 95: (detail) could have named this one `sonar-project-key` for consistency

PR #862, opened 2025-01-09
AAE-30005 Update runtime-version-info.json path

mavotto:
### Checklist

- Jira Reference (also in PR title): https://hyland.atlassian.net/browse/AAE-30005
  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)

### Description
This PR is necessary following the structural refactoring in the hxp-studio-services project.


gionn (changes requested):
see if you can migrate this action to internal-build-tools as it's not really addressing a common use case.

PR #883, opened 2025-01-24
OPSEXP-3066 Use setup-updatecli in the local workflow

gionn:
### Checklist

- Jira Reference (also in PR title): OPSEXP-3066
  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)

### DIFF
--- a/.github/workflows/updatecli.yml
+++ b/.github/workflows/updatecli.yml
@@ -40,7 +40,9 @@ jobs:
         file: ${{fromJson(needs.build-matrix-github-releases.outputs.matrix)}}
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      - uses: updatecli/updatecli-action@304d62420deffb7c824c9cad12942906e16e4352 # v2.76.0
+
+      - name: Install Updatecli
+        uses: ./.github/actions/setup-updatecli
 
       - name: Run Updatecli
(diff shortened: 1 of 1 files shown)


.github/workflows/updatecli.yml
```diff
@@ -40,7 +40,9 @@ jobs:
         file: ${{fromJson(needs.build-matrix-github-releases.outputs.matrix)}}
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      - uses: updatecli/updatecli-action@304d62420deffb7c824c9cad12942906e16e4352 # v2.76.0
+
+      - name: Install Updatecli
+        uses: ./.github/actions/setup-updatecli
```
alxgomz, line 45: Using no ref for an action seems to lead to unexpected behavior. I would assume it should pick the updated ref but doc isn't specific about it https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses.

I guess we can give it a try?

gionn: it's a local workflow (not a reusable one), so should be better use the local ref so we are actually also testing it when it runs

alxgomz: Yeah I got that but I think I miss-read the doc as I first understood not setting the ref could lead to GHA picking a random one... Reading the doc again I wonder why I came up with that conclusion... O_o

alxgomz (approved):

dsibilio (approved):

PR #931, opened 2025-03-05
ACS-9385 Add cache key in setup-java-build

gerardolenski:
### Checklist

- Jira Reference (also in PR title): https://hyland.atlassian.net/browse/ACS-9385
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)
- External PR link where changes has been tested: https://github.com/Alfresco/alfresco-enterprise-repo/pull/2372

### Description

Added the optional possibility to distiguish cache key with custom labeling. It might be useful when one environment cannot access caches created by another. The new parameter is optional, when omitted the previous behaviour of this action hasn't changed.


.github/actions/setup-java-build/action.yml
```diff
@@ -13,16 +13,22 @@ inputs:
...
       uses: actions/cache@d4323d4df104b026a6aa633fdb11d772146be0bf # v4.2.2
+      env:
+        KEY_LABEL: ${{ inputs.cache-key-label != '' && format('-{0}', inputs.cache-key-label) || '' }}
       with:
         path: ~/.m2/repository
-        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        key: ${{ runner.os }}-maven${{ env.KEY_LABEL }}-${{ hashFiles('**/pom.xml') }}
         restore-keys: |
-          ${{ runner.os }}-maven-
+          ${{ runner.os }}-maven${{ env.KEY_LABEL }}-
```
gionn, line 31: you could simplify and just have `maven-` as the default for the input

gerardolenski: Then someone could ovverride it. I wanted to avoid it and have `maven-` always. But I can simplify this in another way. What do you think now?

.github/actions/setup-java-build/action.yml
```diff
@@ -13,16 +13,22 @@ inputs:
     description: the location of the custom Maven settings.xml file to install
     default: ".ci.settings.xml"
     required: false
+  cache-key-label:
+    description: the optional label for the maven repository cache key (eg. to distinguish caches for self hosted runners)
```
gionn, line 17:
```suggestion
    description: the optional label for the maven repository cache key (eg. to handle multiple maven caches within the same repository)
```

.github/actions/setup-java-build/action.yml
```diff
@@ -13,16 +13,22 @@ inputs:
...
+    default: ""
```
gionn, line 18:
```suggestion
    default: "maven"
```

remove `env.KEY_LABEL` completely (don't really seems needed to me), just use `inputs.cache-key-label` directly (altough I will prefer calling it `cache-key-prefix`)

gionn (changes requested):

docs/README.md
```diff
@@ -1562,6 +1565,7 @@ If the Maven settings file is not provided at all, then a default settings file
           java-version: "17" # optional
           java-distribution: "temurin" # optional
           maven-settings: ".ci.settings.xml" # optional
+          cache-key-label: "env1" # optional
```
gionn, line 1568:
```suggestion
          cache-key-prefix: "alternate-maven" # optional
```

version.txt
```diff
@@ -1 +1 @@
-v8.14.1
+v8.15.0
```
gionn, line 1:
```suggestion
v8.16.0
```

gionn, line 1:
```suggestion
v8.16.0
```

gionn (approved):

PR #999, opened 2025-05-06
AAE-34456 Remove camel case from inputs of release-notes-aggregator action

mauriziovitale:
### Checklist

  - [ ] Patch (bugfix)
  - [x] Minor (new feature)


.github/actions/release-notes-aggregator/action.yml
```diff
@@ -1,16 +1,16 @@
 name: Release Notes Aggregator
 description: 'Creates a release notes by aggregating external repositories release notes.'
 inputs:
-  externalRepo:
+  external-repo:
     description: 'The name of the external repositories to aggregate release notes from'
     required: true
-  generateRNfromVersion:
+  generate-rn-from-version:
```
gionn, line 7: from-version and to-version are enough to understand what they are doing in this context

version.txt
```diff
@@ -1 +1 @@
-v8.20.0
+v8.21.0
```
gionn, line 1: it's more a bugfix than an enhancement?

gionn (changes requested):

version.txt
dsibilio: Technically it'd be a breaking change

mauriziovitale: yep is a breaking change, so I don't think a patch if correct

gionn: I know, but this action has been introduced a few days ago and you are the only one using it :D

version.txt
```diff
@@ -1 +1 @@
-v8.20.0
+v8.22.0
```
gionn, line 1:
```suggestion
v8.21.0
```

gionn, line 1:
```suggestion
v8.21.0
```

mauriziovitale: No idea, fixed

gionn (approved):

PR #826, opened 2024-12-04
AAE-28530 Teams notifications improvements

atchertchian:
- move to adaptative cards for send-teams-notification-action: unused input "overwrite" was removed
- add markdown message support for teams usage from reportportal-summarize

### Checklist

- Jira Reference (also in PR title): AAE-28530
- [README](https://github.com/Alfresco/alfresco-build-tools/blob/master/docs/README.md) updated after adding/changing behaviour of an action
- Proposed version increment for [release](https://github.com/Alfresco/alfresco-build-tools/blob/master/docs/README.md#release):
  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)
  - [x] Major (breaking changes)
- External PR link where changes has been tested:

PR #1007, opened 2025-05-12
AAE-33027 Wrap message body text in Teams notifications

dsibilio:
### Checklist

- Jira Reference (also in PR title): https://hyland.atlassian.net/browse/AAE-33027
  - [x] Patch (bugfix)
  - [ ] Minor (new feature)

### Description

Message body should have `wrap: true` otherwise it gets cut off. Tested here: https://teams.microsoft.com/l/message/19:15939c03a8694e8db1b3119ff5ddfe01@thread.tacv2/1747064848775?tenantId=8ca5db88-a5ab-48f7-a5e0-4ce50935f807&groupId=0212f351-7697-4df3-9330-a91abb1f34fa&parentMessageId=1747064848775&teamName=Automate&channelName=%F0%9F%91%B7%F0%9F%8F%BC%20Automate%20Front%20End&createdTime=1747064848775


gionn (approved):

.github/actions/send-teams-notification/action.yml
```diff
@@ -285,7 +285,8 @@ runs:
                             },
                             {
                               "type": "TextBlock",
-                              "text": "${{ env.MESSAGE_OUTPUT }} "
+                              "text": "${{ env.MESSAGE_OUTPUT }} ",
+                              "wrap": true
```
atchertchian, line 289: I looked at the test message but I'm not sure it's cut only because of the missing "wrap": could it be linked to the limit on related bash code? I remember we needed to revisit this number (that came from slack limits) see https://github.com/Alfresco/alfresco-build-tools/pull/826#discussion_r1870932789

This cut can also break pseudo-html/markdown formatting maybe

I'm still not sure if the test was successful given error at https://github.com/Alfresco/hxp-frontend-apps/actions/runs/14976630730/job/42070647647#step:8:531 (?)

dsibilio: 1. It's definitely related to a cut because of missing wrap, you can prove this because if you expand the Teams window you see "more" of the message. If you shrink it, then you see "less" of the message (it gets cut off earlier instead of being wrapped to the next line, to accomodate the window size).
2. The limit for Teams notifications sent via a Webhook should be fairly generous (waaay bigger than the one we had on Slack), The [docs](https://learn.microsoft.com/en-us/microsoftteams/platform/webhooks-and-connectors/how-to/add-incoming-webhook?tabs=newteams%2Cdotnet) mention a 28KB limit per message which is several dozens of thousands worth of characters.
3. The notification job "fails" due to a known error that is unrelated, although the notification comes out successfully: https://github.com/Skitionek/notify-microsoft-teams/pull/99. This can be fixed by bumping the skitionek action to v1.0.9, however this caused other issues with legacy Webhooks (legacy Webhook-related jobs will start to fail in the same way instead as soon as we bump to v1.0.9, pretty much).

For 3., the idea is to migrate all repos from legacy webhooks to workflow-based webhooks and then migrate again to skitionek v1.0.9 and adopt it everywhere, then we should have green notify jobs everywhere 👍

dsibilio: Also discussed with @atchertchian offline, will go ahead and merge this PR as all doubts have been clarified.

PR #1009, opened 2025-05-15
AAE-3308 create actions to build maven projects in parallel

erdemedeiros:
### Checklist

- Jira Reference (also in PR title): AAE-3308
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)
    - used by https://github.com/Alfresco/hxp-process-services/pull/3408
    - PR without preview: https://github.com/Alfresco/hxp-process-services/actions/runs/15145681539
    - PR with preview: https://github.com/Alfresco/hxp-process-services/actions/runs/15143814794/attempts/1
    - Push: https://github.com/Alfresco/hxp-process-services/actions/runs/15142205507
    - changes on existing `maven-build-and-tag`: https://github.com/Alfresco/alfresco-identity-adapter-service/actions/runs/15131853387/job/42534484927


.github/actions/maven-build-and-tag/action.yml
```diff
@@ -117,89 +117,27 @@ runs:
...
       id: update-pom-to-next-version
       if: github.event_name == 'push' || env.IS_PREVIEW == 'true'
       uses: Alfresco/alfresco-build-tools/.github/actions/update-pom-to-next-pre-release@v8.21.1
       with:
         property-to-update: ${{ inputs.property-to-update }}
-        maven-cli-opts: ${{ steps.compute-maven-options.outputs.result }}
-        version: ${{ steps.set-preview-version.outputs.preview-version }}
+        maven-cli-opts: ${{ steps.maven-configure.outputs.maven-options }}
+        version: ${{ steps.maven-configure.outputs.preview-version }}
+        prerelease-type: 'temp'
```
erdemedeiros, line 140: Don't forget to remove before merging

.github/actions/maven-build/action.yml
```diff
@@ -0,0 +1,280 @@
...
+        echo "group-path=$M2_GROUP_FULL_PATH" >> $GITHUB_OUTPUT
+        echo "pattern=$INCLUSION_PATTERN" >> $GITHUB_OUTPUT
+
+    - name: Upload M2 Build artifacts
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+      if: inputs.m2-current-build-upload-name!= ''
+      with:
+        name: ${{ inputs.m2-current-build-upload-name}}
+        path: ${{ steps.m2-inclusion-pattern.outputs.pattern }}
+        retention-days: ${{ inputs.uploaded-artifacts-retention-days }}
```
erdemedeiros, line 253: Not uploading jars files generated by the current build results in sonar goal slowing down because it attempts to download artifacts from remote.

.github/actions/maven-build/action.yml
```diff
@@ -0,0 +1,280 @@
...
+
+    - name: Get project information
+      id: project-info
+      shell: bash
+      run: |
+        VERSION=$(yq e '.project.version' pom.xml)
+        echo "version=$VERSION" >> $GITHUB_OUTPUT
+
+    - name: Set version env variable
+      if: github.event_name == 'push' || env.IS_PREVIEW == 'true'
```
atchertchian, line 139: why do we set this env variable only on push or preview? why do we need it in env?

.github/actions/maven-configure/action.yml
```diff
@@ -0,0 +1,125 @@
...
+      id: set-use-maven_build-cache-env-variable
+      env:
+        MAVEN_BUILD_CACHE_ENABLED: ${{ inputs.maven-build-cache-key != '' && !contains(github.event.pull_request.labels.*.name, inputs.maven-build-cache-skip-label) }}
+        MAVEN_BUILD_CACHE_KEY: ${{ runner.os }}-build-cache-${{ inputs.maven-build-cache-key }}-${{ env.GITHUB_PR_NUMBER || github.ref_name }}-${{ github.run_id }}-${{ github.run_attempt }}
+      shell: bash
+      run: |
+        echo "MAVEN_BUILD_CACHE_ENABLED=$MAVEN_BUILD_CACHE_ENABLED" >> $GITHUB_ENV
+        echo "MAVEN_BUILD_CACHE_KEY=$MAVEN_BUILD_CACHE_KEY" >> $GITHUB_ENV
+
+    - uses: actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
```
atchertchian, line 117: I would consider this part of the build more than configuration

.github/actions/docker-scan-image-dirs/action.yml
```diff
@@ -5,29 +5,68 @@ inputs:
     description: base directory to be scanned
     required: false
     default: .
+  extra-shortener-pattern:
```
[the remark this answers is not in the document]
erdemedeiros: Correct, it's not affecting docker artifacts. It has an impact only in the entry `short-name` in the generated json. By default `short-name` is already keeping only the folder name holding the docker image (instead of the entire path to it), that's why I used "extra": because there is already a transformation in place that is making names shorter anyway, but maybe it's overkill and we can keep only shortener-pattern.
Just to make it clear with an example:
If no pattern it's provided the generated json will look like:
```json
{
    "path": "./some/path/to/my/docker-image/images-my-service",
    "skip-docker": false,
    "short-name": "images-my-service"
  }
```

If the pattern `images-` is provided it becomes:
```json
{
    "path": "./some/path/to/my/docker-image/images-my-service",
    "skip-docker": false,
    "short-name": "my-service"
  }
```

.github/actions/maven-build/action.yml
```diff
@@ -0,0 +1,280 @@
+name: Build Maven project
+description: >-
+  Builds a maven project using the provided command. If the inputs `jacoco-report-name`, `target-folder-upload-name` and `m2-current-build-upload-name` are provided,
+  it also generates aggregated coverage reports and makes them available as build artifact for a next job processing it. It's typically followed by a job containing a step
+  with the action `sonar-scan-on-built-project`.
+inputs:
```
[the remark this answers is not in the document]
erdemedeiros: I haven't experimented any issues related to the number of inputs, even though I have to agree that this action have too many inputs. This is caused mainly because it's aggregating the configuration of java-setup, maven-setup and docker registries login. Maybe we need to re-think how it's done in the future, but for now I tried to follow the same pattern that is already used for the action `maven-build-and-tag` and I wanted to avoid doing a bigger refactoring as part of this PR that's just bringing the possibility of building maven modules as part of matrix.

.github/actions/maven-configure/action.yml
erdemedeiros: I've extracted this from existing build-and-tag that's now reusing this action. If we want to keep backward compatibility we need to keep it.

.github/actions/maven-tag/action.yml
```diff
@@ -0,0 +1,100 @@
+name: Maven Tag
+description: Updates POM files to the next pre-release, commits changes and creates a Git tag.
```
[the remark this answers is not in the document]
erdemedeiros: Yes, the logic differs a bit: 
- maven-build-and-tag:
    - update pom
    - build project
    - create git tag
- maven-tag:
    - update pom
    - create build tag
    
So it's not possible to reuse it without changing the current behavior.

.github/actions/docker-scan-image-dirs/action.yml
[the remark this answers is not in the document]
erdemedeiros: Ok, go for `name-shortener-pattern`

atchertchian (approved):

revatilimaye (approved):

PR #1015, opened 2025-05-20
AAE-34998 Add action to check PR metadata on push/pr

mauriziovitale:
### Checklist

- Jira Reference (also in PR title): AAE-34998
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)
- External PR link where changes has been tested: https://github.com/Activiti/Activiti/pull/5020


.github/actions/dependabot-check/action.yml
```diff
@@ -0,0 +1,34 @@
...
+        SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+        FOUND_LABEL: ${{ inputs.label }}
+      shell: bash
+      run: |
+        LABELS=$(gh api /repos/$GITHUB_REPOSITORY/commits/$SHA/pulls | jq -r '.[].labels.[].name')
+        echo "Labels found: $LABELS"
+        if [[ "$GITHUB_ACTOR" == "dependabot[bot]" && "$LABELS" == *"$FOUND_LABEL"* ]]
+        then
+          echo "PR opened by dependabot and with gh_action. The result is true"
+          echo "SKIP=true" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
mauriziovitale: You don’t know what u are talking about

atchertchian (commented):
Changes look good to me but I wonder if we want another name for the action, so I will let @gionn and/or @dsibilio share their thoughts (github-dependabot-check? (?))

.github/actions/github-dependabot-check/action.yml
```diff
@@ -0,0 +1,43 @@
+name: Dependabot check
+description: Checks if author is dependabot and if specific label is added to PR
+
+inputs:
+  gh-token:
+    description: 'GitHub token'
```
dsibilio, line 6: Would be good to expand the description here explaining which permissions are expected for the provided GitHub token to work correctly in the context of this action.

.github/actions/github-dependabot-check/action.yml
```diff
@@ -0,0 +1,43 @@
...
+    required: true
+  label:
```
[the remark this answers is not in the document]
mauriziovitale: we can do this one will be and if needed :)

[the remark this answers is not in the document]
mauriziovitale: I don't see it can be applied right now. The only use case we are interested in is checking if 1 label added automatically by dependabot is present and those are exclusive
gh_action
docker
maven
npm

docs/README.md
```diff
@@ -740,6 +741,19 @@ With proper concurrency logic in place, the latest run might have been cancelled
...
+Action to check if a pull request is opened by dependabot with a specific label.
+
+```yaml
+      - name: dependabot check
+        id: dependabot
+        uses: ./.github/actions/github-dependabot-check
+        with:
+          gh-token: ${{ secrets.GITHUB_TOKEN }}
+          label: github_actions
+```
```
gionn, line 755: would be cool to have a more comprehensive/real world example because right now I am struggling to see how much reusable this is (how do you actually use the result of this action?)

gionn (changes requested):

docs/README.md
mauriziovitale: https://github.com/Activiti/Activiti/pull/5020/files#diff-1578e1a2c827dc71148f1362dddb7cab08c67fe01518fb0f198fc3b571e6cba7R30

@gionn  this is how we are going to use it.. basically we need an action to identify if a PR was opened by dependabot with a specific label gh_action to avoid the deploy or any other CI operation not needed to save time

mauriziovitale: @gionn  I've added a more real use case. let me know

docs/README.md
```diff
@@ -740,6 +741,43 @@ With proper concurrency logic in place, the latest run might have been cancelled
           workflow: my-workflow.yml
 ```
 
+### github-dependabot-check
+
+Action to check if a pull request is opened by dependabot with a specific label.
+
+```yaml
+      name: CI
+
```
dsibilio, line 750:
```suggestion
```

dsibilio: Unnecessary part of the example, we can get rid of it

docs/README.md
```diff
@@ -740,6 +741,43 @@ With proper concurrency logic in place, the latest run might have been cancelled
...
+      - name: dependabot check
+        id: dependabot
+        uses: ./.github/actions/github-dependabot-check
+        with:
+          gh-token: ${{ secrets.GITHUB_TOKEN }}
+          label: github_actions
+
+  deploy:
+    needs: check-dependabot
+    if: steps.dependabot.outputs.is_dependabot != 'true'
```
dsibilio, line 767:
```suggestion
    if: steps.dependabot.outputs.result != 'true'
```

Isn't this called `result`?

docs/README.md
```diff
@@ -740,6 +741,43 @@ With proper concurrency logic in place, the latest run might have been cancelled
...
+    steps:
+      - name: Deploy
+        run: echo "Deploying..."
+```
+How it works
+The dependabot check step uses your action to check if the PR is from Dependabot and has the github_actions label.
+If both conditions are true, the action sets an output (e.g., is_dependabot: true).
+The deploy job only runs if is_dependabot is not true, so deploys are skipped for these PRs.
+Benefit:
+This saves CI/CD resources and time by skipping unnecessary deploys for automated dependency updates that only affect GitHub Actions workflows.
```
dsibilio, line 778:
```suggestion
How it works
The dependabot check step uses your action to check if the PR is from Dependabot and has the github_actions label.
If both conditions are true, the action sets an output (e.g., result: true).
The deploy job only runs if result is not true, so deploys are skipped for these PRs.
Benefit:
This saves CI/CD resources and time by skipping unnecessary deploys for automated dependency updates that only affect GitHub Actions workflows.
```

Paragraph mentions `is_dependabot` instead of `result`. Also formatting is off and it shows all on one line, needs more new lines.

docs/README.md
mauriziovitale: the only reason why I have been asked to move the action in this public is to reuse it across the 8 or more BE automate repos. If the action is only returning the metadata I don't see the point to have an action I can do the call in each repo instead of referencing the action.

gionn: What if Dependabot PRs aren't the only ones you want to skip? Other automations might use github-action[bot] or a custom GitHub user. And what if you're merging a user-created PR that shouldn't trigger propagation because the changes don't need it?
What if you have a complex workflow for which you could have a different outcome than just skip the job?

Same desired behavior, but very different triggering conditions.

mauriziovitale: right now this is a very specific scenario we would like to tackle. 
dependabot + gh_action 
once we will collect different data and cases we can change/improve the action. 
Again, initially I was creating the action as part of the repo but I was told to export it as shared since we have several BE repo that are suffering from this. 
On FE we use the Nx tool that allows us to handle different scenarios.

.github/actions/github-dependabot-check/action.yml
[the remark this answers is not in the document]
atchertchian: We can always add a "labels" input later if needed - abd the action is easy enough to use several times so that people can do OR/AND custom logic with each result

docs/README.md
atchertchian: Indeed we would like to tackle this and see if it really helps us with unneeded propagations. 
I like what you're suggesting @gionn and I agree it would be much more generic, but I don't want us to have to do hardcode that user is dependabot, and we can use this action several times if we want to check more than one label (and do the AND/OR custom logic) on the caller side.
If we made the name more generic, and introduced an "actor" input that would default to `dependabot[bot]`, do you think that could be enough as a first step?

docs/README.md
atchertchian: I reviewed the text and fixed the markdown formatting that pre-commit was also complaining about - please review @mauriziovitale @dsibilio

docs/README.md
atchertchian: Reading the README, I even wonder if it would make sense to have github_actions label by default too.

Also our alternative is pushing to Activiti/Activiti actions for easier reuse everywhere if that's too specific (I did that once for [a similar need](https://github.com/Activiti/Activiti/pull/4395)), we can go back to that even if it's not the cleanest for our version references

.github/actions/github-dependabot-check/action.yml
```diff
@@ -0,0 +1,44 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Check
+      id: check
+      env:
+        GH_TOKEN: ${{ inputs.gh-token }}
+        SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+        FOUND_LABEL: ${{ inputs.label }}
+        GITHUB_EVENT_NAME: ${{ github.event_name }}
```
atchertchian, line 26: this one is already mapped in bash, it should not need to be redefined (like GITHUB_ACTOR) https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

docs/README.md
atchertchian: here is a suggestion (better name for the action to be found): https://github.com/Alfresco/alfresco-build-tools/pull/1015/commits/c290b059231e3003a52c9130b4c8904eb12e8f7e

atchertchian: my understanding is that the action logic makes it work both on push (after PR merge) and on pull_request, right @mauriziovitale? even if we'll use it on push only for our use case - I have other use cases I see now (to detect propagation PRs for instance) where this could be reused

dsibilio: While it can technically be used on `pull_request` I would probably discourage using it? In case of a `pull_request` event I'd assume we'd already have both the author and the labels as part of the event and it's fairly trivial to extract it & run conditional logic based on it without having to make a gh api call 🤔

mauriziovitale: you should always fetch the label from api if you want to get the latest status and not only rely on those originally added when the PR was created. Btw I am fine with any solution.

atchertchian: Yes as Maurizio says we've been fetching labels from API anyway as it's more reliable

`github-pr-check`?

dsibilio: `github-pr-check-metadata` sounds good to me

.github/actions/github-dependabot-check/action.yml
```diff
@@ -0,0 +1,50 @@
+name: PR author and label check
+description: Checks PR author and label, using PR that led to merge on push
+
+inputs:
+  gh-token:
+    description: 'GitHub token with content read permission'
+    required: true
+  actor:
+    description: 'Check if the pull request was created by this actor'
+    required: false
```
gionn, line 10: maybe you want to add dependabot as a default if dependabot is the main/initial use case?

then if you rename the action to something more generic it's fine for me to merge

docs/README.md
atchertchian: Rename done

PR #1042, opened 2025-06-18
AAE-35226 Add new actions to handle Dependabot PRs

atchertchian:
### Checklist

  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)
  - [X] Major (breaking changes)
- External PR link where changes has been tested: TODO

### Description

This introduces new actions to:
- gracefully fail the workflow early when it's not going to run ok (replacing the logic at https://github.com/Activiti/Activiti/blob/8.8.0-alpha.62/.github/actions/check-ext-build/action.yml)
- facilitate validating and merging Dependabot PR in a single click, requiring only one human approval

An old related action that is not following good practices anyway was also removed.


docs/README.md
```diff
@@ -49,13 +48,15 @@ Here follows the list of GitHub Actions topics available in the current document
...
     - [git-latest-tag](#git-latest-tag)
+    - [github-check-secrets-source](#github-check-secrets-source)
     - [github-check-upcoming-runs](#github-check-upcoming-runs)
     - [github-deployment-create](#github-deployment-create)
     - [github-deployment-status-update](#github-deployment-status-update)
     - [github-deployments-delete](#github-deployments-delete)
     - [github-download-file](#github-download-file)
     - [github-https-auth](#github-https-auth)
     - [github-list-changes](#github-list-changes)
+    - [github-manage-dependabot-pr-approval](#github-manage-dependabot-pr-approval)
```
atchertchian, line 59: Any naming suggestion is welcome

.github/workflows/test.yml
```diff
@@ -4,16 +4,19 @@ on:
...
-      types: [ submitted ]
 
 jobs:
   build:
     # Trigger a run when approving a bot PR or as usual
     if: >-
-      (github.event.review.state == 'approved' && contains(fromJSON('["dependabot[bot]", "github-actions[bot]"]'), github.event.pull_request.user.login)) ||
-      (github.actor != 'dependabot[bot]' && github.event_name != 'pull_request_review')
+      (github.event.review.state == 'approved' && contains(fromJSON('["atchertchian", "github-actions[bot]"]'), github.event.pull_request.user.login)) ||
+      (github.actor != 'atchertchian' && github.event_name != 'pull_request_review')
```
atchertchian, line 19: I did this blindly just to test (this is a temporary change, see [tmp] commit) but I'm not actually sure we should skip on dependabot PRs: maybe we should actually fail (using the check-actions-source new action?)

Similarly maybe we want to align the dependabot process on this repo and trigger on milestone (and maybe setup auto-merge on approval (?))

Thoughts welcome @gionn @dsibilio

.github/actions/github-check-secrets-source/action.yml
```diff
@@ -0,0 +1,19 @@
...
+        SECRET_SOURCE: ${{ github.secret_source }}
+      shell: bash
+      run: |
+        if [[ $SECRET_SOURCE == 'Dependabot' ]]
+        then
+          echo "::error::Access to secrets forbidden to Dependabot: follow your repository process for PR validation" && exit 1
+        fi
+        if [[ $SECRET_SOURCE != 'Actions' && $SECRET_SOURCE != 'Codespaces' ]]
+        then
+          echo "::error::Missing proper secrets source: failing build" && exit 1
```
gionn, line 18:
```suggestion
          echo "::error::Unexpected $SECRET_SOURCE secrets source: failing build" && exit 1
```

.github/actions/github-check-secrets-source/action.yml
```diff
@@ -0,0 +1,19 @@
+name: Check Github Actions Source
+description: Fails if actions source is not valid for secrets usage
```
gionn, line 2: A good name should tell the purpose and not the implementation, so if here we are triggering an early failure could be something like `fail-fast-with-dependabot`?

anyway is this really different from just checking `github.actor`?

why do you prefer failing the build in this case? for us always worked well to add a condition at job level to skip it (as described in the README) so you won't even billed for this.

.github/actions/github-manage-dependabot-pr-approval/action.yml
```diff
@@ -0,0 +1,81 @@
...
+      shell: bash
+      run: |
+        if [[ -n "$MILESTONE" ]]; then
+          echo "::notice::Milestone PR to trigger PR validation: $PR_URL"
+          gh pr edit -R $GITHUB_REPOSITORY $PR_NUMBER --milestone "$MILESTONE"
+        fi
+        if [[ -n "$LABEL" ]]; then
+          echo "::notice::Label PR to trigger PR validation: $PR_URL"
+          gh pr edit -R $GITHUB_REPOSITORY $PR_NUMBER --add-label "$LABEL"
+        fi
```
gionn, line 59: Hold on, if the goal here is to fully automate Dependabot PR updates, you're introducing a supply chain attack risk. Automated merging of untrusted dependency updates can expose GitHub Actions workflows to malicious code execution, potentially leaking secrets. This is precisely why Dependabot uses its own restricted secrets context, separate from the standard github secrets used in regular workflows.

Dependabot PRs should be reviewed by a human before triggering any workflows that require secrets. A reviewer can then approve or manually trigger a build with the necessary permissions.

The only alternative is to design your workflows to avoid using secrets entirely, allowing builds to run in a reduced-privilege mode. This way, you can still catch potential issues introduced by the dependency bump, though the coverage will be limited.

cc @ciairuzz

gionn (changes requested):

.github/actions/github-manage-dependabot-pr-approval/action.yml
gionn: I understood only now this is meant to be used in a workflow triggered by an actual approval, sorry for the noise 🙏🏻

docs/README.md
gionn: could not even mention dependabot if the actor could become an input

.github/actions/github-check-secrets-source/action.yml
atchertchian: the problem with skips is that it can easily lead the PR validation to ignore required checks: we prefer failing all the builds where we need secrets for proper validation

.github/actions/github-manage-dependabot-pr-approval/action.yml
atchertchian: I did check with @ciairuzz if it was ok to use a PAT to trigger rerun, and indeed since we can trace it back to approval we're ok

docs/README.md
```diff
@@ -848,6 +857,32 @@ This action requires a checkout with `fetch-depth: 0` option as follow:
...
+      - submitted
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Alfresco/alfresco-build-tools/.github/actions/github-pr-manage-dependabot-approval@ref
+        with:
+          github-token: ${{ secrets.BOT_GITHUB_TOKEN }}
+          milestone-on-approval: Dependabot
```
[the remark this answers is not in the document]
atchertchian: thanks let me check what we actually need to replicate

.github/workflows/test.yml
atchertchian: will test on another repo (will revert temp changes)

gionn (approved):

.github/actions/github-trigger-approved-pr/action.yml
```diff
@@ -0,0 +1,86 @@
...
+          echo "::notice::Milestone PR to trigger PR validation: $PR_URL"
+          gh pr edit -R $GITHUB_REPOSITORY $PR_NUMBER --milestone "$MILESTONE"
+        fi
+        if [[ -n "$LABEL" ]]; then
+          echo "::notice::Label PR to trigger PR validation: $PR_URL"
+          gh pr edit -R $GITHUB_REPOSITORY $PR_NUMBER --add-label "$LABEL"
+        fi
+
+    - uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+      if: steps.check.outputs.continue == 'true' and inputs.auto-merge-on-approval == 'true'
```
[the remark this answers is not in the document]
atchertchian: thanks!

wojciech-piotrowiak (approved):

gionn (approved):
Overall it looks good and I understand the different use cases. It may be worth adding an item in the cookbook section to describe the approach of using these actions together, since looking at them individually may not fully reveal their purpose.

.github/actions/github-trigger-approved-pr/action.yml
```diff
@@ -0,0 +1,88 @@
...
+        if [[ -n "$MILESTONE" ]]; then
+          echo "::notice::Milestone PR to trigger PR validation: $PR_URL"
+          gh pr edit -R $GITHUB_REPOSITORY $PR_NUMBER --milestone "$MILESTONE"
+        fi
+        if [[ -n "$LABEL" ]]; then
+          echo "::notice::Label PR to trigger PR validation: $PR_URL"
+          gh pr edit -R $GITHUB_REPOSITORY $PR_NUMBER --add-label "$LABEL"
+        fi
+
+    - uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
```
dsibilio, line 68:
```suggestion
    - uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
```

dsibilio (approved):

wojciech-piotrowiak (approved):

PR #1044, opened 2025-06-23
First update to try to silent process

Sourav-Bhattacharyya:
### Checklist

  - [ ] Patch (bugfix)
  - [x] Minor (new feature)

### Description

Trying to silence tag lines

### DIFF
--- a/.github/actions/calculate-next-internal-version/next-prerelease.sh
+++ b/.github/actions/calculate-next-internal-version/next-prerelease.sh
@@ -7,7 +7,7 @@ then
 fi
 
 echo "Fetching tags ... "
-git fetch --tags
+git fetch --tags --no-recurse-submodules --quiet
 
 FIRST_PRERELEASE_SUFFIX="-${PRERELEASE_TYPE}.1"
 


.github/workflows/test.yml
```diff
@@ -78,6 +81,7 @@ jobs:
       - uses: ./.github/actions/setup-pysemver
       - uses: ./.github/actions/calculate-next-internal-version
         with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
atchertchian, line 84: this would should not be needed anymore with the default expression (to be tested)

Sourav-Bhattacharyya: Works fine updated it in my recent commit.

.github/actions/maven-tag/action.yml
```diff
@@ -52,12 +55,14 @@ runs:
...
       uses: Alfresco/alfresco-build-tools/.github/actions/update-pom-to-next-pre-release@v8.24.1
       with:
+        github-token: ${{ inputs.github-token }}
         property-to-update: ${{ inputs.property-to-update }}
         maven-cli-opts: ${{ steps.maven-configure.outputs.maven-options }}
         prerelease-type: ${{ inputs.prerelease-type }}
       env:
         MAVEN_USERNAME: ${{ inputs.maven-username }}
         MAVEN_PASSWORD: ${{ inputs.maven-password }}
+        GITHUB_TOKEN: ${{ inputs.github-token }}
```
atchertchian, line 65: you should not need both as there's a precendence system, you don't want to override default env variables resolution

Sourav-Bhattacharyya: Done

.github/actions/calculate-next-internal-version/action.yml
```diff
@@ -22,9 +25,19 @@ runs:
     - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
     - uses: Alfresco/alfresco-build-tools/.github/actions/setup-pysemver@v8.24.1
     - id: next-prerelease-resolver
-      run: ${{ github.action_path }}/next-prerelease.sh
       shell: bash
+      run: |
+        if [[ -n "${{ inputs.github-token }}" ]]; then
```
atchertchian, line 30: It seems you went for a backward compatible strategy after all :+1: 
Don't forget not to require the input anymore if so

.github/actions/calculate-next-internal-version/next-prerelease.sh
```diff
@@ -1,24 +1,51 @@
...
-echo "Fetching tags ... "
-git fetch --tags
+API_URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/tags"
+PER_PAGE=100
+PAGE=1
+MATCHING_TAG=""
+echo "::group::Fetching tags from GitHub API"
+
+while true; do
+  response=$(curl -s -H "Authorization: Bearer ${GITHUB_TOKEN}" \
```
atchertchian, line 22: (nitpick info: the gh cli should be available on gh runners to help doing such calls in a more user-friendly way, I'm not asking you to change it, but please consider it for next time or if you think it'll improve maintainability of this code)

.github/actions/maven-build-and-tag/action.yml
```diff
@@ -66,6 +66,9 @@ inputs:
   git-username:
     description: The username to commit on the git repository
     required: true
+  github-token:
+    description: "GitHub token to use for API requests"
+    required: true
```
atchertchian, line 71: this also should not be required after all, for compat?

.github/actions/release-notes-aggregator/dist/index.js
```diff
@@ -31457,23 +31457,23 @@ module.exports = parseParams
...
 /******/ 	// Load entry module and return exports
 /******/ 	// This entry module is referenced by other modules so it can't be inlined
 /******/ 	var __webpack_exports__ = __nccwpck_require__(399);
 /******/ 	module.exports = __webpack_exports__;
-/******/ 	
+/******/
 /******/ })()
-;
\ No newline at end of file
+;
```
atchertchian, line 31479: I'm pretty sure dist folders are excluded from formatting for a reason, and this does not related to your changes. I would suggest to commit independently (with proper formatting if we want to enforce this on dist folders content too)

atchertchian (commented):
I left a few comments below (mostly to ensure backward compatibility) - please check if some bats tests could be added for all these bash scripts, thanks!

.github/actions/calculate-next-internal-version/action.yml
atchertchian: that matches the copilot feedback number 2 (suppressed because of low confidence!)

.github/actions/calculate-next-internal-version/next-prerelease.sh
```diff
@@ -1,24 +1,51 @@
...
+  for tag in $tag_names; do
+    if [[ "$tag" =~ ^${NEXT_VERSION}-${PRERELEASE_TYPE}\.[0-9]+$ ]]; then
+      echo "Matched: $tag"
+      MATCHING_TAG="$tag"
+      break 2
+    fi
+  done
 
-echo "Next version: $NEXT_VERSION"
-LATEST_PRERELEASE="$(git tag --sort=-creatordate | grep -m 1  "^$NEXT_VERSION\-$PRERELEASE_TYPE\.[[:digit:]]\{1,4\}$" | cat)"
```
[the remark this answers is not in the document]
atchertchian: cc @erdemedeiros

.github/actions/calculate-next-internal-version/test/next-prerelease.bats
```diff
@@ -0,0 +1,65 @@
...
+  done <<< "$tag_names"
+
+  count=$(echo "$tag_names" | grep -c . || true)
+  if (( count < PER_PAGE )); then
+    break
+  fi
+
+  PAGE=$((PAGE + 1))
+done
+echo "::endgroup::"
```
erdemedeiros, line 45: @Sourav-Bhattacharyya , @souvikpanda99 , in the end, my feeling is that by using directly the Github API is bringing a lot of complexity. Specially because we need to handle pagination: I was hopping we could filter directly by a given pattern in the API call and get them order properly, but it seems Githup api does not support that.
In addition, maintaining two versions of it (plain git + github API) makes things even more complex.
Based on that, I'd suggest we proceed as the following:

1. Revert back to the old logic (plain git) and only add `--quiet` option to avoid flooding the logs ([KISS principle](https://blog.bit.ai/software-design-principles/))
2. add the addition output with the latest pre-release to avoid calculating it again in other workflows (as mentioned by @atchertchian )
Does it make sense?

erdemedeiros (changes requested):

.github/actions/calculate-next-internal-version/test/next-prerelease.bats
atchertchian: I agree that if the problem was the noise, the optimized remote retrieval of tags might not actually be more optimized, as calling git fetch --tags on a gh actions runner should not be that long even on a repo with a lot of tags

atchertchian (approved):

.github/actions/release-notes-aggregator/dist/index.js
erdemedeiros: @Sourav-Bhattacharyya maybe you can revert the changes in the file

erdemedeiros (approved):
LGTM. You can bump the [version](https://github.com/Alfresco/alfresco-build-tools/blob/master/version.txt) to `v8.25.0` to new release containing your change once it's merged.

.github/actions/calculate-next-internal-version/next-prerelease.sh
```diff
@@ -16,9 +16,11 @@ LATEST_PRERELEASE="$(git tag --sort=-creatordate | grep -m 1  "^$NEXT_VERSION\-$
 if [ -n "$LATEST_PRERELEASE" ]; then
   echo "Latest prerelease version found: $LATEST_PRERELEASE"
   NEXT_PRERELEASE="$(pysemver bump prerelease "$LATEST_PRERELEASE")"
+  echo "latest-prerelease=$LATEST_PRERELEASE" >> $GITHUB_OUTPUT
 else
   echo "No prerelease found for version $NEXT_VERSION yet"
   NEXT_PRERELEASE="$NEXT_VERSION$FIRST_PRERELEASE_SUFFIX"
+  echo "latest-prerelease=" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
Sourav-Bhattacharyya: Can I resolve this and the other conversations?

.github/actions/release-notes-aggregator/dist/index.js
Sourav-Bhattacharyya: Done

erdemedeiros (approved):

PR #1053, opened 2025-07-03
Fix release workflow

gionn:
### Description

Partially revert https://github.com/Alfresco/alfresco-build-tools/commit/c7762a338bb357d021e74c923a85a4fd6aa3eed6


.github/workflows/release.yml
```diff
@@ -38,6 +38,7 @@ jobs:
 
           - uses: stefanzweifel/git-auto-commit-action@778341af668090896ca464160c2def5d1d1a3eb0 # v6.0.1
             with:
+              branch: master
```
atchertchian, line 41: copilot says it should be "main" :)

PR #1061, opened 2025-07-17
AAE-36650 Add dependabot-coverage-check action to validate GitHub Action

wojciech-piotrowiak:
…ions in dependabot.yml

### Checklist

  - [ ] Patch (bugfix)
  - [X] Minor (new feature)

### DIFF
--- a/.github/actions/dependabot-coverage-check/action.yml
+++ b/.github/actions/dependabot-coverage-check/action.yml
@@ -1,0 +1,32 @@
+name: dependabot-coverage-check
+author: wojciech.piotrowiak
+description: Workaround for https://github.com/dependabot/dependabot-core/issues/6345. This action checks if all GitHub Actions in .github/actions are listed in .github/dependabot.yml.
+runs:
+  using: composite
+
+  steps:
+    - name: Checkout code
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+    - name: Fail if missing dependabot config
+      shell: bash
+      run: |
+        # List all action directories
+        actions=$(find .github/actions -mindepth 1 -maxdepth 1 -type d | sed 's|^./||')
+
+        # Extract listed directories from dependabot.yml
+        listed=$(yq '.updates[] | select(.package-ecosystem == "github-actions") | .directories[]' .github/dependabot.yml | sed 's|^/||' | grep '^.github/actions/')
+
+        # Check for missing actions
+        missing=""
+        for action in $actions; do
+        if ! echo "$listed" | grep -qx "$action"; then
+        missing="$missing\n$action"
+        fi
+        done
+
+        if [ -n "$missing" ]; then
+        echo -e "Missing actions in dependabot.yml:$missing"
+        exit 1
+        else
+        echo "All actions are listed in dependabot.yml"
+        fi
--- a/docs/README.md
+++ b/docs/README.md
@@ -33,6 +33,7 @@ Here follows the list of GitHub Actions topics available in the current document
     - [automate-propagation](#automate-propagation)
     - [calculate-next-internal-version](#calculate-next-internal-version)
     - [configure-git-author](#configure-git-author)
+    - [dependabot-coverage-check](#dependabot-coverage-check)
     - [dbp-charts](#dbp-charts)
     - [dispatch-resume-workflow](#dispatch-resume-workflow)
     - [docker-build-image](#docker-build-image)
@@ -420,6 +421,11 @@ Configures the git username and email to associate commits with the provided ide
 ```
 
 The two vars in the previous snippet are [workflow configuration variables](https://github.blog/changelog/2023-01-10-github-actions-support-for-configuration-variables-in-workflows/) that can be created at organization level and shared across different repositories.
+
+### dependabot-coverage-check
+
+This action checks if all GitHub Actions are listed in the `.github/dependabot.yml` file.
+This is a workaround for a [GH Link](https://github.com/dependabot/dependabot-core/issues/6345) problem where global configuration does not work.
 
 ### dbp-charts
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -82,6 +82,14 @@ updates:
         patterns:
           - "*"
   - package-ecosystem: "github-actions"
+    directory: "/.github/actions/dependabot-coverage-check"
+    schedule:
+      interval: "weekly"
+    groups:
+      catch-all:
+        patterns:
+          - "*"
+  - package-ecosystem: "github-actions"
     directory: "/.github/actions/dispatch-resume-workflow"
     schedule:
       interval: "weekly"


.github/actions/dependabot-coverage-check/action.yml
```diff
@@ -0,0 +1,32 @@
...
+        # List all action directories
+        actions=$(find .github/actions -mindepth 1 -maxdepth 1 -type d | sed 's|^./||')
+
+        # Extract listed directories from dependabot.yml
+        listed=$(yq '.updates[] | select(.package-ecosystem == "github-actions") | .directories[]' .github/dependabot.yml | sed 's|^/||' | grep '^.github/actions/')
+
+        # Check for missing actions
+        missing=""
+        for action in $actions; do
+        if ! echo "$listed" | grep -qx "$action"; then
```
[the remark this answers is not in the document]
atchertchian: @wojciech-piotrowiak do you have a test run somewhere please?

.github/actions/dependabot-coverage-check/action.yml
```diff
@@ -0,0 +1,32 @@
+name: dependabot-coverage-check
```
atchertchian, line 1: It would be nice to call this action in the local test.yml workflow: the templating system is supposed to take care of it already during pre-commit, but it can be seen as a test of this action, wdyt?

.github/actions/dependabot-coverage-check/action.yml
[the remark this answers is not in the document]
wojciech-piotrowiak: Test added

.github/actions/dependabot-coverage-check/dependabot-coverage-check.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+# dependabot-coverage-check
+# Author: wojciech.piotrowiak
+# Description: Workaround for https://github.com/dependabot/dependabot-core/issues/6345.
+# This script checks if all GitHub Actions in .github/actions are listed in .github/dependabot.yml.
```
gionn, line 6: update to keep them consistent

anyway I am not so inclined to duplicate these info here given they are already provided in the action.yml

.github/actions/dependabot-coverage-check/dependabot-coverage-check.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Find repository root (where .git directory exists)
```
gionn, line 8: please drop all of these LLM inflated comments which just state the obvious

.github/actions/dependabot-coverage-check/dependabot-coverage-check.sh
```diff
@@ -0,0 +1,52 @@
...
+  exit 1
+fi
+
+if [ ! -f ".github/dependabot.yml" ]; then
+  echo "Error: .github/dependabot.yml file not found"
+  exit 1
+fi
+
+# Find all directories that contain action.yml files
+actions=$(find .github/actions -name "action.yml" -type f | xargs -I {} dirname {} | sed 's|^./||')
```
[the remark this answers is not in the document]
gionn: ensure there is a test for this edge case, if it fails then do what it suggest :D

.github/actions/dependabot-coverage-check/dependabot-coverage-check.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Extract listed directories from dependabot.yml - support both directory and directories
+listed_single=$(yq '.updates[] | select(.package-ecosystem == "github-actions") | .directory' .github/dependabot.yml | sed 's|^/||' | grep '^\.github/actions/' | sed 's|^\./||')
```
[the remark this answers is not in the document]
gionn: add a command -v check at the beginning of the script

.pre-commit-config.yaml
```diff
@@ -38,3 +38,7 @@ repos:
         language: system
         entry: ./check_readme.sh
         pass_filenames: false
+      - id: check-dependabot-coverage
+        name: Check if all actions are listed in dependabot.yml
+        language: system
+        entry: .github/actions/dependabot-coverage-check/dependabot-coverage-check.sh
```
gionn, line 44: not sure it really make sense here given that we already have `generate-dependabot` before which will never make this to trigger.

.github/actions/dependabot-coverage-check/action.yml
gionn: there are bats tests, there is really no need to add it elsewhere

atchertchian: I meant to call the action from the .github/workflows/test.yml file (after pre-commit, as we're advertizing the action in this repo, not the pre-commit bash script, no? (I mean it could also be documented as is if that's what we want to do).
That's also why I ws saying this action can added to the list of actions that are being called on the common pre-checks actiion in the other "build-tools" repo: this way we do not need to impact all pre-commit configurations for this extra "common" check, we just need to bump the action version. wdyt?

.github/actions/dependabot-coverage-check/action.yml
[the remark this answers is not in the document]
atchertchian: As people prefer :)

.pre-commit-config.yaml
atchertchian: that's what I meant in another comment: the action is called, and if it fails, maybe one of those calls is buggy so it's nice to have this in a test.yml pipeline, no?

.github/actions/dependabot-coverage-check/action.yml
atchertchian: also note that adding bash scripts to .pre-commit local conf is sometimes an issue depending on the developer OS

wojciech-piotrowiak: Added that action to test.yml

.github/actions/dependabot-coverage-check/dependabot-coverage-check.sh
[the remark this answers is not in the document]
wojciech-piotrowiak: Done

.github/actions/dependabot-coverage-check/action.yml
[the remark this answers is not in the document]
wojciech-piotrowiak: Action renamed

.github/actions/dependabot-coverage-check/action.yml
```diff
@@ -0,0 +1,12 @@
...
+author: wojciech.piotrowiak
+description: Workaround for https://github.com/dependabot/dependabot-core/issues/6345. This action checks if all GitHub Actions in .github/actions are listed in .github/dependabot.yml.
```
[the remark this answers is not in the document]
wojciech-piotrowiak: Done

.github/actions/dependabot-coverage-check/dependabot-coverage-check.sh
wojciech-piotrowiak: Done

.github/actions/dependabot-missing-actions-check/action.yml
```diff
@@ -0,0 +1,12 @@
+name: dependabot-coverage-check
+author: wojciech.piotrowiak
+description: Ensure that all GitHub Actions defined in the current repo have managed dependabot updates. Workaround for https://github.com/dependabot/dependabot-core/issues/6345
+runs:
+  using: composite
+
+  steps:
+    - name: Checkout code
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
atchertchian, line 9: I would leave that out because most of the time the code will already be checked out on the expected branch/ref to check, we don't want to have to add a boolean input to disable checkout like we do on some actions doing checkout unnecessarily

```suggestion
```

docs/README.md
```diff
@@ -421,6 +422,11 @@ Configures the git username and email to associate commits with the provided ide
 
 The two vars in the previous snippet are [workflow configuration variables](https://github.blog/changelog/2023-01-10-github-actions-support-for-configuration-variables-in-workflows/) that can be created at organization level and shared across different repositories.
 
+### dependabot-missing-actions-check
+
+This action checks if all GitHub Actions are listed in the `.github/dependabot.yml` file.
+This is a workaround for a [GH Link](https://github.com/dependabot/dependabot-core/issues/6345) problem where global configuration does not work.
+
 ### dbp-charts
```
atchertchian, line 430: you're missing a change to version.txt to do a minor release

.github/workflows/test.yml
```diff
@@ -18,6 +18,8 @@ jobs:
         with:
           allowlist: |
             Alfresco/alfresco-build-tools/
+      - name: Dependabot missing actions check
+        uses: ./.github/actions/dependabot-missing-actions-check
```
gionn, line 22: I will move it to the test job below to make it clear it's really just yet another additional test for it

docs/README.md
wojciech-piotrowiak: done

.github/workflows/test.yml
wojciech-piotrowiak: Good point, fixed

PR #1074, opened 2025-08-07
OPSEXP-3377 Add action for image tag computing

pmacius:
### Checklist

- Jira Reference (also in PR title): OPSEXP-3377
- [README](https://github.com/Alfresco/alfresco-build-tools/blob/master/docs/README.md) updated after adding/changing behaviour of an action :white_check_mark:
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)
- External PR link where changes has been tested: https://github.com/Alfresco/devops-performance-testing/pull/2


gionn (changes requested):
there is the upstream https://github.com/docker/metadata-action which address most of these needs, will create a task to better document it here

PR #1120, opened 2025-09-02
AAE-36127 Define pre-commit hook to check sha pinned gh actions versions

atchertchian:
### Checklist

  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -52,6 +52,7 @@ Here follows the list of GitHub Actions topics available in the current document
     - [get-commit-message](#get-commit-message)
     - [git-commit-changes](#git-commit-changes)
     - [git-latest-tag](#git-latest-tag)
+    - [github-actions-ensure-sha-pinned-actions](#github-actions-ensure-sha-pinned-actions)
     - [github-check-upcoming-runs](#github-check-upcoming-runs)
     - [github-deployment-create](#github-deployment-create)
     - [github-deployment-status-update](#github-deployment-status-update)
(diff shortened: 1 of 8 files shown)


.github/actions/pre-commit/action.yml
```diff
@@ -38,6 +38,12 @@ runs:
       with:
         python-version: ${{ inputs.python-version }}
 
+    - name: Install Python dependencies for local hooks
+      shell: bash
+      run: |
+        python -m pip install --upgrade pip
+        pip install requests urllib3
```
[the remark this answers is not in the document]
gionn: and more importantly we don't want them installed on every pre-commit action invocation

try defining `additional_dependencies` at the hook configuration level

.pre-commit-hooks.yaml
```diff
@@ -28,6 +28,13 @@
   entry: bash -c 'find . -name Chart.yaml | xargs -I% bash -c "dirname %" | xargs -t -I% bash -c "helm template % | kubepug --error-on-deprecated --error-on-deleted --input-file=-"'
   language: system
   pass_filenames: false
+- id: gha-sha-convert
+  name: GitHub Actions SHA Converter
+  description: Convert GitHub Actions to use SHA-pinned versions
+  entry: python3 gha-sha-convert/gha_sha_convert.py
+  language: system
```
gionn, line 35: tell me again why can't we use python here

.pre-commit-config.yaml
```diff
@@ -42,3 +42,10 @@ repos:
         language: system
         entry: ./check_readme.sh
         pass_filenames: false
+      - id: gha-sha-convert
+        name: GitHub Actions SHA Converter
+        description: Convert GitHub Actions to use SHA-pinned versions
+        entry: python3 gha-sha-convert/gha_sha_convert.py
+        language: system
+        files: ^\.github/(workflows|actions)/.*\.(yml|yaml)$
+        pass_filenames: true
```
gionn, line 51: it feels ugly but I think it would be better if we self-reference this repo here with a release tag so we avoid duplicating configs and we ensure it's really working as expected

which means this won't run the hook if changed in a branch but maybe it's a feature

docs/README.md
```diff
@@ -737,6 +738,68 @@ Gets the latest tag and commit sha for the given pattern. The result is returned
           pattern: 1.0.0-alpha*
 ```
 
+### github-actions-ensure-sha-pinned-actions
```
gionn, line 741: this is duplicating the docs inside the action?

docs/security.md
```diff
@@ -11,11 +11,51 @@ Before creating / modifying any GitHub Actions workflow make sure you're familia
...
   - [Secrets detection](#secrets-detection)
     - [First setup](#first-setup)
     - [Updating new/old secrets to the baseline](#updating-newold-secrets-to-the-baseline)
       - [Excluding multiple secrets via regex](#excluding-multiple-secrets-via-regex)
 
+## SHA Pinning for GitHub Actions
+
+To enhance supply chain security, it's crucial to pin GitHub Actions to specific SHA hashes instead of using mutable tag references. This prevents potential supply chain attacks where malicious code could be injected by compromising action repositories.
+
+### Why SHA Pinning Matters
```
gionn, line 24: again no duplicated docs here, I think here we just need to have a link to the action (this section was more to educate about the security approach rather than documenting security related actions)

gha-sha-convert.sh
gionn, line 1: just for reference - delete before merging

gha-sha-convert/.pre-commit-hooks.yaml
gionn, line 1: this is not really working right?

docs/README.md
atchertchian: Yes in case people want to use this without the pre commit tooling boilerplate (?)

PR #1199, opened 2025-11-14
AAE-40011 Add custom property to skip also unit tests

ffazzini:
### Checklist

  - [x] Patch (bugfix)
  - [ ] Minor (new feature)

### Description

We introduced a custom property in order to allow us to skip either surefire or failsafe tests plugins so that we can run unit or integration tests separately. Note that skipTests is default for both plugins.

After introducing this property the behaviour of adding the skip-tests label had a side effect, the PR creator now has to wait for couple of mins for the unit tests to run (as skipUTs replaced the skipTests in surefire).  

It is not a big problem to wait for few mins, also there is a new PR created by Elias which will build the images in parallel so the PR creator does not have to wait 2-3mins for the unit tests when using the skip-tests label in the PR.
So we could either merge this PR and fix this problem or ignore this PR for the reasons mentioned above.

### DIFF
--- a/.github/actions/maven-configure/action.yml
+++ b/.github/actions/maven-configure/action.yml
@@ -76,7 +76,7 @@ runs:
         NTP='--no-transfer-progress'
         ${{ inputs.verbose }} && NTP=''
         TEST_OPTIONS=''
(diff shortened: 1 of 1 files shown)


jsokolowskii (approved):

gionn (approved):
you are not bumping version.txt for a release but we can release one once we have merged the weekly automated bumps

PR #1414, opened 2026-05-27
AAE-46385 Add Supply Chain Review flow and github-upsert-comment action

dsibilio:
### Checklist

- Jira Reference (also in PR title): AAE-46385
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)
- External PR link where changes has been tested: https://github.com/HylandSoftware/supply-chain-guardian-demo/pull/33

### Description

- Adding the first GitHub Agentic Workflow to alfresco-build-tools.
- Adding a supporting reusable workflow to create/update PR comments based on an identifier, that is useful to share context about which commands are available to trigger which GitHub Agentic Workflow, as a common approach is to trigger them via a `/slash_command` as a PR comment.

Tested here: https://github.com/HylandSoftware/supply-chain-guardian-demo/pull/33#issuecomment-4555618817


.pre-commit-config.yaml
```diff
@@ -12,27 +12,29 @@ repos:
...
-        exclude: '.*/dist/.*'
+        exclude: '(.*/dist/.*|\.github/workflows/.*\.lock\.yml$)'
 
   - repo: https://github.com/sirosen/check-jsonschema
     rev: 0.37.1
     hooks:
       - id: check-dependabot
       - id: check-github-actions
       - id: check-github-workflows
+        exclude: ^\.github/workflows/.*\.lock\.yml$
```
atchertchian, line 25: (just a question, I understand and agree with adding this exclude) Is this check actually failing on actual lock wotkflows?

docs/README.md
```diff
@@ -2679,6 +2683,127 @@ Release command has access to the following environment variables:
 See the dedicated [terraform](terraform.md) for more information on the reusable
 workflows provided by us.
 
+### add-pr-comment
```
atchertchian: I'd prefer "create-or-update" to "add", and I would find it more useful to have this exposed as an action than workflow for reusability 🤔

.github/workflows/supply-chain-review.md
```diff
@@ -0,0 +1,343 @@
...
+
+### 2b. OpenSSF Scorecard
+
+Fetch the project health score. First determine the source repository from the package metadata (see 2c), then query:
+
+`GET https://api.scorecard.dev/projects/github.com/{owner}/{repo}`
+
+Record the overall score (0-10) and individual check results. Pay special attention to: Maintained, Code-Review, Vulnerabilities, Branch-Protection, Signed-Releases. A score below 3 is very concerning; below 5 warrants caution. If the package has no linked source repository, the scorecard will be unavailable — this is itself a risk signal.
+
+### 2c. Package Metadata (Registry)
```
atchertchian, line 93: Do we need instructions here for more registries, or do we have specific instructions here because they differ from standard? (thinking about terraform, github-actions, .python libs ..)

.github/workflows/supply-chain-review.md
```diff
@@ -0,0 +1,343 @@
...
+- Old version (or mark as `NEW DEPENDENCY` if newly added)
+- New version
+
+If no dependency files were changed, post a brief PR comment stating that no dependency changes were detected and no review is needed, then stop.
+
+## Step 1b — Filter Internal Dependencies
+
+Before collecting external data, identify and exclude internal/private dependencies that cannot be resolved by public APIs.
+
+**Internal dependency namespaces (skip these):**
```
atchertchian, line 59: Said differently, we do not perform transitive analysis, and this will not detect any issues in internal dependencies, we assume that those should be covered by a similar automated detection process, right?

.github/workflows/supply-chain-review.md
```diff
@@ -0,0 +1,343 @@
...
+
+These are guidelines, not rigid formulas. Use judgment to combine signals — multiple medium signals can compound into a high-risk assessment.
+
+**Risk level thresholds**: 0-20 = LOW, 21-50 = MEDIUM, 51-80 = HIGH, 81-100 = CRITICAL.
+
+If no suspicious patterns are found, assign a score of 0-10 and risk level LOW. Most legitimate upgrades should score LOW.
+
+## Step 5 — Post Findings as PR Comment
+
+Before posting, check if an existing PR comment from this workflow already exists by looking for a comment whose body starts with `## Supply Chain Security Review`. If one exists, update that comment with the new report. If none exists, add a new comment. This avoids cluttering the PR with duplicate reports on re-runs.
```
atchertchian: why not have the previous comment deleted and a new one created? (just a wonder for where comment feedback will be in PR timeline/history)

atchertchian (commented):
(LGTM, mostly nitpicks)

.pre-commit-config.yaml
dsibilio: Yes, it would. I limited myself at the minimum number of possible exclusions that would have otherwise made `pre-commit run --all-files` fail when running against the `.lock.yml` files which do not follow all the conventions of traditional flows.

docs/README.md
dsibilio: Sure, we can do that.

.github/workflows/supply-chain-review.md
dsibilio: We would likely need additional instructions and firewall rules to check also for Terraform/GHA/Python lib bumps.

My initial intention was to have a first release supporting NPM/Maven ecosystems only.

.github/workflows/supply-chain-review.md
dsibilio: To have an internal dependency bumped you can imagine that someone else made a PR in the provider repo that produces the artifact for the internal dependency, likely changing the `pom.xml` e.g. bumping/refactoring libraries there, and that would be the phase where we expect the reminder to perform a `/supply-chain-review` would appear to the developer.

The flow tries to sometimes investigate autonomously transitive dependencies when it makes sense already (e.g. adding a new version of a library may provide older versions of transitive deps causing more CVEs to pop up in case they're closer to the root/not governed by dependencyManagement) FYI, see e.g.: https://github.com/HylandSoftware/supply-chain-guardian-demo/pull/10#issuecomment-4553351299.

```txt
The upstream POM does pin very old transitive versions (jackson-databind:2.4.1 with 48 OSV advisories; httpclient:4.2.6 with 3 advisories), but Maven resolution in this application does not keep those versions. The effective tree for this PR resolves:

    com.fasterxml.jackson.core:jackson-databind:2.13.4.2
    org.apache.httpcomponents:httpclient:4.5.13
    org.apache.httpcomponents:httpmime:4.5.13
    org.codehaus.woodstox:woodstox-core-asl:4.2.0

That materially lowers the immediate vulnerability risk versus the package's own published POM.
```

In general I wouldn't enhance the instructions here to force the flow to always analyze transitive dependencies as some libraries have so many it would likely make the flow very slow and much more expensive.

.github/workflows/supply-chain-review.md
dsibilio: I forgot we even had this instruction in the prompt as it does not work, I'll reword it.

I don't want to give this flow the permission to delete PR comments, but I want to test the option to `hide-older-comments: true` on `add-comments` outputs to see if that works for us!

docs/README.md
dsibilio: Moved to internal action `github-upsert-comment` @atchertchian

.github/workflows/supply-chain-review.md
dsibilio: Agreed offline to start supporting these ecosystems initially and see if worth expanding later on 👍

.github/workflows/supply-chain-review.md
dsibilio: Removed from the instructions and added the hint to hide older comments, seems to work great, see e.g.:
- https://github.com/HylandSoftware/supply-chain-guardian-demo/pull/33
- https://github.com/HylandSoftware/supply-chain-guardian-demo/pull/9

katarzynakita (approved):

mikolajbrzezinski (approved):

docs/README.md
```diff
@@ -2679,6 +2723,96 @@ Release command has access to the following environment variables:
...
+### General Introduction
+
+GitHub Agentic Workflows are a new type of workflow that leverage the GitHub Copilot engine to perform complex tasks that require reasoning, decision-making, and interaction with external APIs. These workflows are designed to automate tasks that are difficult to achieve with traditional workflows, such as code reviews, security analysis, and project management.
+
+#### For Providers
+
+To update and release a new version of an Agentic Workflow, you can simply edit the md workflow file in this repository, run:
+
+```bash
+gh extension upgrade aw && gh aw compile
```
pmacius, line 2737: Not sure if possible and for sure not something for this PR but maybe in the future we can introduce some pre-commit hook for this

pmacius (approved):

docs/README.md
dsibilio: Thanks for the review @pmacius, I agree that this might be worth exploring as we add more GitHub Agentic Workflows to this repo. Will keep it in mind 🙏

PR #1415, opened 2026-05-27
Added Import GPG Key action, added a smoke test and updated README

adecatur72:
### Checklist

  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)


pmacius (commented):
I found https://github.com/crazy-max/ghaction-import-gpg does it suit your use case? If yes then maybe we can use it and add documentation here about it.

alxgomz (commented):
If meant to be re-useable I would avoid appending lines in the default gpg file to avoid breaking upstream env.
We should either test the presence options to avoid dups or IIRC gpg can be given non default directory to work in so we are sure we do not mess up with anything (could be an opt-in option)

alxgomz (approved):

PR #1446, opened 2026-06-30
OPSEXP-3938 Add setup-maven action

pmacius:
### Checklist

- Jira Reference (also in PR title): OPSEXP-3938
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)

### Description

Why

  We need a reliable, pinned way to provision a chosen Maven version on runners, independent of whatever Maven the runner image happens to ship.

  How it works

  - Input version (required) — the Maven version to install, e.g. 3.9.9. Validated against <major>.<minor>.<patch>; an invalid value fails fast
  with a clear error.
  - Download + integrity check — fetches apache-maven-<version>-bin.tar.gz from archive.apache.org and verifies its SHA-512 checksum before
  extracting. Refuses to proceed on mismatch.
  - Install location — extracts under $RUNNER_TEMP/apache-maven/<version>, exports MAVEN_HOME, and prepends bin/ to the PATH.
  - Caching — caches the distribution via actions/cache, keyed on apache-maven-<os>-<arch>-<version>, so it isn't re-downloaded on every run. On
  a cache hit the download/verify is skipped.
  - Input cache-key-suffix (optional) — appended to the cache key to disambiguate parallel jobs that would otherwise share a key, or to force a
  fresh cache.
  - Verify — runs mvn --version at the end as a sanity check that the install and PATH wiring took effect.

  Tests

  - New test-setup-maven helper action asserts the resolved mvn version matches the requested one and that MAVEN_HOME points at the installed
  version.
  - test.yml exercises: install 3.9.9, install 3.8.8, and rejection of an invalid version string.

  Docs

  - Added a setup-maven section to docs/README.md with a usage example.

### DIFF
--- a/.github/actions/setup-maven/action.yml
+++ b/.github/actions/setup-maven/action.yml
@@ -2,9 +2,8 @@ name: 'Setup Maven'
 description: 'Install a specific Apache Maven version and add it to the PATH.'
 inputs:
   version:
-    description: 'Version of Apache Maven to install (e.g. 3.9.9). If empty, uses the Maven version already on the PATH.'
-    required: false
-    default: ''
+    description: 'Version of Apache Maven to install (e.g. 3.9.9).'
+    required: true
 outputs:
   version:
     description: 'Version of Maven that is available on the PATH after this action runs.'
@@ -12,8 +11,8 @@ outputs:
 runs:
   using: "composite"
   steps:
-    - name: Install Maven
-      if: inputs.version != ''
+    - name: Resolve install paths
+      id: paths
       shell: bash
       env:
         MAVEN_VERSION: ${{ inputs.version }}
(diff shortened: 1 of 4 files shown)


.github/actions/setup-maven/action.yml
```diff
@@ -0,0 +1,82 @@
+name: 'Setup Maven'
+description: 'Install a specific Apache Maven version and add it to the PATH.'
+inputs:
+  version:
+    description: 'Version of Apache Maven to install (e.g. 3.9.9). If empty, uses the Maven version already on the PATH.'
+    required: false
+    default: ''
+outputs:
```
gionn: I am confused by this empty default, if I want to use whatever maven version is available on PATH I would never call this action - which is for setting up a custom version :D

.github/actions/setup-maven/action.yml
```diff
@@ -0,0 +1,82 @@
...
+      run: |
+        set -euo pipefail
+
+        if [[ ! "$MAVEN_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+          echo "::error::Invalid Maven version '$MAVEN_VERSION'. Expected format: <major>.<minor>.<patch> (e.g. 3.9.9)."
+          exit 1
+        fi
+
+        MAJOR="${MAVEN_VERSION%%.*}"
+        INSTALL_DIR="$RUNNER_TOOL_CACHE/apache-maven/$MAVEN_VERSION"
```
gionn: I am not sure RUNNER_TOOL_CACHE is meant to be used in this way, it's where github actions expose the available tooling, but why you can't just use a temp folder instead of this?

gionn (changes requested):
you need to add actions/cache in order to avoid downloading it every single time

.github/actions/setup-maven/action.yml
pmacius: True its better to use temp

docs/README.md
```diff
@@ -2446,6 +2447,21 @@ Install the Kubernetes preupgrade checker and add it to the PATH.
           version: "1.3.2"
 ```
 
+### setup-maven
+
+Install a specific Apache Maven version from the Apache archive and add it to
+the PATH. The distribution is cached across runs (via `actions/cache`, keyed on
+the OS, architecture and requested version) to avoid re-downloading it every
+time. Set `cache-key-suffix` to disambiguate parallel jobs that would otherwise
```
pmacius, line 2455: "disambiguate" :bowtie:

.github/actions/setup-maven/action.yml
```diff
@@ -0,0 +1,107 @@
...
+
+        VERSION=$(mvn -v | awk '/^Apache Maven/ {print $3; exit}')
+        echo "Detected Maven version on PATH: $VERSION"
+
+        if [[ "$VERSION" != "$REQUESTED_VERSION" ]]; then
+          echo "::error::Maven on PATH ($VERSION) does not match the requested version ($REQUESTED_VERSION). PATH updates may not have taken effect; check that no later step overwrites \$PATH or \$MAVEN_HOME."
+          exit 1
+        fi
+
+        echo "version=$VERSION" >> "$GITHUB_OUTPUT"
```
gionn: this feels a bit too slop

can you just run `mvn --version` which should be enough to cover everything?

having an output for a value which I give in input doesn't really feel something useful

pmacius: Yes thats right

gionn (approved):
remember to raise another pr to replace any existing workflow/action with this new action

PR #1472, opened 2026-07-13
OPSEXP-4107 support custom card actions and body

pmacius:
Add optional card-actions and card-extra-body inputs so callers can append Adaptive Card action buttons (e.g. Action.OpenUrl) and extra body elements (FactSet, monospace log blocks) to the notification card. Both default to an empty array, so existing usages render unchanged.

### Checklist

- Jira Reference (also in PR title): OPSEXP-4107
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)
- External PR link where changes has been tested: https://github.com/Alfresco/terraform-alfresco-pipeline/pull/1951


gionn (commented):
can you try replacing skitionek/notify-microsoft-teams with a simple curl? I see we are always using `raw` and then passing just the webhook url and dry_run, that action doesn't feel doing anything more useful than just an http request

gionn (approved):

PR #1476, opened 2026-07-21
ACS-12079 Add maven-compute-release-versions composite action

damianujma:
### Checklist

- Jira Reference (also in PR title): https://hyland.atlassian.net/browse/ACS-12079
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)
- https://github.com/Alfresco/alfresco-community-repo/actions/runs/29840031364/job/88666264306
- https://github.com/Alfresco/alfresco-community-repo/pull/4221

### Description

This PR it to add a new `maven-compute-release-versions` composite action that derives `release-version` and `development-version` from the current POM `SNAPSHOT` version.
Action strips the `-SNAPSHOT` suffix for the release version and increments the patch segment for the next development version.

The action reads the POM version via `mvn help:evaluate`, then computes:
  - `release-version`: e.g. `26.2.0-SNAPSHOT` → `26.2.0`
  - `development-version`: e.g. `26.2.0` → `26.2.1-SNAPSHOT`

### DIFF
--- a/.github/actions/maven-compute-release-versions/action.yml
+++ b/.github/actions/maven-compute-release-versions/action.yml
@@ -1,7 +1,7 @@
 name: Maven Compute Release Versions
 description: >
-  Derives the release version and next development version from the current POM SNAPSHOT version.
-  Strips the -SNAPSHOT suffix to produce the release version, then increments the patch segment
+  Derives the release version and next development version from the current POM version (typically a -SNAPSHOT version).
+  Strips a trailing -SNAPSHOT suffix to produce the release version, then increments the last numeric segment
   for the next development version. Maven must be available on PATH before invoking this action.
 
 outputs:
--- a/.github/actions/maven-compute-release-versions/action.yml
+++ b/.github/actions/maven-compute-release-versions/action.yml
@@ -1,0 +1,20 @@
+name: Maven Compute Release Versions
+description: >
+  Derives the release version and next development version from the current POM SNAPSHOT version.
+  Strips the -SNAPSHOT suffix to produce the release version, then increments the patch segment
+  for the next development version. Maven must be available on PATH before invoking this action.
+
+outputs:
+  release-version:
+    description: "The release version, e.g. 26.2.0"
+    value: ${{ steps.compute.outputs.release-version }}
+  development-version:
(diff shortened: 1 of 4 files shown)


.github/actions/maven-compute-release-versions/action.yml
```diff
@@ -0,0 +1,20 @@
...
+description: >
+  Derives the release version and next development version from the current POM SNAPSHOT version.
+  Strips the -SNAPSHOT suffix to produce the release version, then increments the patch segment
+  for the next development version. Maven must be available on PATH before invoking this action.
+
+outputs:
+  release-version:
+    description: "The release version, e.g. 26.2.0"
+    value: ${{ steps.compute.outputs.release-version }}
+  development-version:
```
gionn: next-version sounds more clear

.github/actions/maven-compute-release-versions/action.yml
```diff
@@ -0,0 +1,20 @@
+name: Maven Compute Release Versions
+description: >
+  Derives the release version and next development version from the current POM SNAPSHOT version.
+  Strips the -SNAPSHOT suffix to produce the release version, then increments the patch segment
```
gionn: is your current convention to always increment the patch version? Most development work adds features or non-breaking changes, so bumping the minor version is the more common case. Patch increments should be reserved for pure bugfix releases (e.g. from a support branch)

anyway could be something worth having it handled via an input

.github/actions/maven-compute-release-versions/action.yml
damianujma: Renamed to next-development-version.

.github/actions/maven-compute-release-versions/action.yml
damianujma: Yes. There is a group of repositories, such as alfresco-community-repo and alfresco-tas-utility, that always require a patch version increment after each commit. In most cases, however, increasing the minor version should be a deliberate decision, so I don't see a strong reason to expose it as an input parameter now.

gionn (commented):
check copilot review comments

gionn (approved):

kmagdziarz (approved):

PR #1485, opened 2026-07-28
Require new actions to justify reusability in copilot instructions

gionn:
### Description

New actions were being added to `alfresco-build-tools` as one-off extractions from a single workflow rather than genuinely shared building blocks (e.g. `get-downstream-commit-message` in #1484). This adds an explicit reusability requirement to `.github/copilot-instructions.md`, in both the Action Development Guidelines and the Pull Request Checklist, stating that new actions must solve a problem multiple callers realistically need, not just look generic on paper.

This is an internal instructions-only change with no effect on any action's behavior, so no README update or version-increment label applies.


alxgomz (approved):
👍

PR #1491, opened 2026-07-31
OPSEXP-4208 Filter extracted Jira tickets to the configured project before setting fix version

gionn:
### Checklist

- Jira Reference (also in PR title): OPSEXP-4208
  - [x] Patch (bugfix)
  - [ ] Minor (new feature)

### Description

`jira-propagate-release` extracts Jira ticket keys from the GitHub release payload (name, tag, URL, body) with a project-agnostic regex, then passes all of them to `jira-set-fix-version` to stamp the release's Fix Version. `jira-set-fix-version` only supports a single project per run (it derives the project from the first issue key and enforces all issues belong to it), and the Fix Version is only ever created in the single project configured via `jira-project-key`. When the release notes happened to reference a ticket from a different Jira project (e.g. `ACS-11432` alongside the intended `OPSEXP-*` tickets), the whole step hard-failed because that version doesn't exist in the unrelated project, even though the intended tickets would have updated fine. This is what broke https://github.com/Alfresco/acs-deployment/actions/runs/30633986769/job/91166856127.

This PR filters the extracted ticket list down to only tickets belonging to the configured `jira-project-key` before handing them to `jira-set-fix-version`, via a new optional `JIRA_PROJECT_KEY` env var on `extract-tickets-from-release-payload.sh` (wired from the existing `jira-project-key` action input, no new inputs added). Added bats coverage for the filtering behavior and confirmed it's backward compatible when the key is unset.

OPSEXP-4208

### DIFF
--- a/.github/actions/jira-propagate-release/extract-tickets-from-release-payload.sh
+++ b/.github/actions/jira-propagate-release/extract-tickets-from-release-payload.sh
@@ -11,6 +11,7 @@ set -euo pipefail
 #   GITHUB_OUTPUT                 If set, write outputs there (GitHub Actions outputs file)
 #   GITHUB_VERSION_PREFIX         Prefix to remove from GitHub tag (literal match)
 #   JIRA_VERSION_PREFIX           Prefix to prepend to Jira version name
+#   JIRA_PROJECT_KEY              If set, keep only extracted tickets belonging to this Jira project (e.g. "ABC")
 #
 # Output keys:
 #   tickets-csv=ABC-1,DEF-2        Comma-separated list of unique ticket IDs (empty if none found)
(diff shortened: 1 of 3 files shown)


.github/actions/jira-propagate-release/extract-tickets-from-release-payload.sh
```diff
@@ -89,6 +90,10 @@ if [[ ${grep_status} -eq 2 ]]; then
   exit 1
 fi
 
+if [[ -n "${JIRA_PROJECT_KEY:-}" && -n "${MATCHES}" ]]; then
```
atchertchian, line 93: the extraction script would be nice to share in dedicated action, is it specific to release format or could we consider it works also on a commit message (we do copy release notes when promoting some changes)?

gionn: extraction logic is just grepping a regex against a text payload so not hard to replicate, most of the code here is to ensure we are properly handling an event payload for a github release (which is what this action expect)

pmacius (approved):

PR #1515, opened 2026-08-18
OPSEXP-4294 fix(env-load-from-yaml): treat values as literals and tolerate missing files

gionn:
### Checklist

- Jira Reference (also in PR title): OPSEXP-4294
  - [x] Patch (bugfix)
  - [ ] Minor (new feature)
- External PR link where changes has been tested: https://github.com/Alfresco/terraform-alfresco-pipeline/pull/1992

### Description

The `env-load-from-yaml` rewrite in v18.19.1 (#1475) parsed each value with `eval "value=\"$rhs\""`, which re-split quoted multi-word values on spaces (breaking things like `MAVEN_CLI_OPTS="--batch-mode --quiet -U"`, whose second word then ran as a command), stripped quotes out of JSON values, and could even execute backtick/`$(...)` sequences embedded in a value since both are expanded inside double quotes by bash. It also turned a missing `yml_path` into a hard failure, breaking callers that point it at an optional per-environment override file.

This replaces the `eval`-based parsing with `envsubst`, so a value is taken verbatim from after `=` to end of line (quotes, spaces, JSON braces/brackets preserved byte-for-byte) with only `$VAR`/`${VAR}` expansion applied, and makes a missing `yml_path` a no-op (with a `::warning::` logged) instead of an error.

Adds bats coverage for literal quoting, JSON round-tripping, `$VAR` expansion inside an otherwise-literal value, the missing-file no-op, and regression tests proving `$(...)`/backtick payloads are written literally and never executed.

### DIFF
--- a/.github/actions/env-load-from-yaml/env-load-from-yaml.sh
+++ b/.github/actions/env-load-from-yaml/env-load-from-yaml.sh
@@ -1,12 +1,11 @@
 #!/bin/bash
 set -eo pipefail
 
-# A missing input file is a no-op: callers rely on being able to point this
-# at an optional per-environment override file.
-[ -f "$YML_PATH" ] || exit 0
+if [ ! -f "$YML_PATH" ]; then
+  echo "::warning::$YML_PATH not found, skipping"
+  exit 0
+fi
 
-# Capture yq output before looping so a yq failure aborts the script
-# instead of being swallowed by process substitution.
 entries="$(yq '.env.global[]' "$YML_PATH")"
 
 while IFS= read -r ENVVAR; do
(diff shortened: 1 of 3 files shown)


.github/actions/env-load-from-yaml/env-load-from-yaml.sh
```diff
@@ -1,8 +1,12 @@
 #!/bin/bash
 set -eo pipefail
 
-# Capture yq output before looping so a yq failure/missing file aborts the
-# script instead of being swallowed by process substitution.
+# A missing input file is a no-op: callers rely on being able to point this
+# at an optional per-environment override file.
+[ -f "$YML_PATH" ] || exit 0
```
gionn: add a warning and then exit.

in the comment don't explain why a specific caller needs this

.github/actions/env-load-from-yaml/env-load-from-yaml.sh
```diff
@@ -1,8 +1,12 @@
...
+
+# Capture yq output before looping so a yq failure aborts the script
+# instead of being swallowed by process substitution.
```
gionn:
```suggestion
```

pmacius (approved):
LGTM

alxgomz (approved):

PR #1517, opened 2026-08-19
AAE-50050 Restore commit message variable expansion in git-commit-changes

mkrbr:
### Checklist

- Jira Reference (also in PR title): https://hyland.atlassian.net/browse/AAE-50050
  - [x] Patch (bugfix)
  - [ ] Minor (new feature)

### Description
image

### DIFF
--- a/.github/actions/git-commit-changes/action.yml
+++ b/.github/actions/git-commit-changes/action.yml
@@ -44,7 +44,6 @@ runs:
         GIT_COMMIT_USERNAME: ${{ inputs.username }}
         GIT_COMMIT_EMAIL: ${{ inputs.email }}
         GIT_COMMIT_ADD_OPTIONS: ${{ inputs.add-options }}
-        GIT_COMMIT_MESSAGE: ${{ inputs.commit-message }}
         GIT_COMMIT_OPTIONS: ${{ inputs.commit-options }}
         GIT_COMMIT_REPO_DIR: ${{ inputs.repository-directory }}
         GIT_COMMIT_SKIP_IF_NO_CHANGES: ${{ inputs.skip-if-no-changes }}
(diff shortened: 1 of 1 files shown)


JWollweber:
theoretically a injection vector but better than the one before

JWollweber (approved):

.github/actions/git-commit-changes/action.yml
```diff
@@ -60,6 +60,7 @@ runs:
         git add $GIT_COMMIT_ADD_OPTIONS
         git status
         git --no-pager diff --cached
+        GIT_COMMIT_MESSAGE="$(printf '%s' "$GIT_COMMIT_MESSAGE" | envsubst)"
```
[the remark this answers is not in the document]
mkrbr: we dont use those runners

dsibilio (commented):
I'm not sure I understand why the git commit message can't be passed as already resolved/fully expanded by the consumer, instead of resolving env vars in this action?

.github/actions/git-commit-changes/action.yml
gionn, line 63: it will be trivial to leak GPG_PRIVATE_KEY just by referencing it in a commit message then

gionn (changes requested):

.github/actions/git-commit-changes/action.yml
mkrbr: Yep it can. it's an improved revert. I'm fine to create a separate ticket to resolve the env vars on action usage level and close this pr and we live in the mean time that we are producing messages like this, as it never should've worked this way in the first place:
image
And we close this PR, do you agree?

gionn: in general, it would help having more context in the pr description.

if it's just a matter of replacing $VERSION you can pass arguments to envsubst tp restrict which env will be substituted if present
https://man7.org/linux/man-pages/man1/envsubst.1.html

also check what Domenico said, I see now the commit message is an input of the action so the additional logic can just be on the caller

anyway this pr as it is can't be merged

PR #1522, opened 2026-08-24
OPSEXP-4300 Detect concurrent commits during maven-release-slim

gionn:
### Checklist

- Jira Reference (also in PR title): OPSEXP-4300
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)

### Description

Adds a new `git-branch-drift` action that asserts a branch's remote head still matches an expected SHA, to detect commits pushed to it since a run started. Wires it into `maven-release-slim` (new `fail-on-concurrent-commits` input, default `'true'`), checked once before any work starts and again right after the Maven deploy, so a concurrent merge to `master` during a release aborts the run instead of pushing the release commits onto an unexpected base.

The `test/local-actions` label is needed because `maven-release-slim` references the new action by SHA pin, which does not exist on `master` yet until this PR is released.

This is an advisory check, not a lock — a race window remains between the check and the push. A stronger fix would be a fast-forward-only commit mutation (e.g. an `expectedHeadOid`-style guard on the underlying commit action) and/or a workflow-level `concurrency` group in the caller; both are out of scope here.

OPSEXP-4300


.github/actions/git-branch-drift/action.yml
```diff
@@ -0,0 +1,41 @@
...
+    required: false
+    default: ''
+  branch:
+    description: Branch to check on the remote. Defaults to the current pull request head ref, or the current ref.
+    required: false
+    default: ${{ github.event.pull_request.head.ref || github.ref_name }}
+  remote:
+    description: Git remote to check.
+    required: false
+    default: origin
```
gionn, line 15: probably never needed on github actions

.github/workflows/test.yml
```diff
@@ -333,3 +333,42 @@ jobs:
       - name: Show current branches
         run: |
           echo "Current branch is ${{ steps.branch-info.outputs.branch-name }}"
+
+      #region Test git-check-branch-head
+      - name: Get current remote head for git-check-branch-head test
+        id: remote-head
+        run: echo "sha=$(git ls-remote origin "refs/heads/${{ steps.branch-info.outputs.branch-name }}" | cut -f1)" >> "$GITHUB_OUTPUT"
```
gionn: better to rely on default behaviour?

.github/actions/git-branch-drift/action.yml
gionn: Dropped it, thanks.

.github/actions/git-check-branch-head/git-check-branch-head.sh
```diff
@@ -0,0 +1,33 @@
...
+  echo "Branch '$BRANCH' on '$REMOTE' is still at expected SHA $EXPECTED_SHA"
+  echo "changed=false" >> "$GITHUB_OUTPUT"
+  exit 0
+fi
+
+echo "changed=true" >> "$GITHUB_OUTPUT"
+echo "::error::Branch '$BRANCH' on '$REMOTE' moved from expected $EXPECTED_SHA to $REMOTE_SHA since this run started. New commits landed concurrently; re-run once it is safe."
+if [ "$FAIL_ON_MISMATCH" = "true" ]; then
+  exit 1
+fi
```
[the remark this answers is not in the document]
gionn: Good catch, fixed: the annotation is now plain error only when it actually fails the step, and just a log line otherwise.

.github/actions/git-check-branch-head/action.yml
```diff
@@ -0,0 +1,41 @@
...
+  branch:
+    description: Branch to check on the remote. Defaults to the current pull request head ref, or the current ref.
+    required: false
+    default: ${{ github.event.pull_request.head.ref || github.ref_name }}
+  remote:
+    description: Git remote to check.
+    required: false
+    default: origin
+  repository-directory:
+    description: Optional path to the the directory holding the git repository (defaults to current directory)
```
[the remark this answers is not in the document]
gionn: Fixed by removing the input entirely (see other comment).

.github/workflows/test.yml
gionn: Simplified to rely on the action's own defaults for both expected-sha and branch.

PR #1523, opened 2026-08-26
OPSEXP-4351 Speed up free-hosted-runner-disk-space folder removal

gionn:
### Checklist

- Jira Reference (also in PR title): OPSEXP-4351
  - [ ] Patch (bugfix)
  - [x] Minor (new feature)
  - #1532
  - #1533

### Description

The `free-hosted-runner-disk-space` action now removes cached tools and dependencies in the background: each folder is deleted in parallel rather than one at a time, and the per-folder `du -sh` size report only runs when `diagnose-top-offenders-enabled` is `true`, since walking the directory tree twice has no benefit otherwise. Each background removal's PID is tracked, so the step fails if any of them errors instead of just checking the last one. The tools-cache directory is removed the same way as everything else, in full; GitHub's toolcache library recreates it on demand. Since the tools-cache directory contains the CodeQL cache as a subdirectory, its size report tolerates files that vanish mid-walk while the CodeQL folder is being deleted in parallel.

### Benchmark

Ran the `Free up disk space on the runner` step 10 times on master (#1532) and 10 times on this branch (#1533) using `gh run rerun` on the `test (ubuntu-latest)` job. `diagnose-top-offenders-enabled` was forced to `false` on both sides, since the diagnostic `du` walk would otherwise mask the real gain.

| Run | master | this PR | diff |
|---|---|---|---|
| 1 | 176s | 125s | +51s |
| 2 | 55s | 67s | −12s |
| 3 | 93s | 67s | +26s |
| 4 | 109s | 24s | +85s |
| 5 | 87s | 28s | +59s |
| 6 | 59s | 69s | −10s |
| 7 | 86s | 65s | +21s |
| 8 | 113s | 111s | +2s |
| 9 | 137s | 75s | +62s |
| 10 | 198s | 28s | +170s |
| **avg** | **111.3s** | **65.9s** | **+45.4s** |
| **median** | **101s** | **67s** | **+34s** |

This branch was faster in 8 of 10 runs, cutting average time by about 41% and median by about 34%. The two regressions were small (12s and 10s) next to wins of up to 170s. Master's times also swung much more (55s to 198s) than this branch's (24s to 125s), which matches what backgrounding the removals should do: total time is now bounded by the slowest single deletion instead of the sum of all of them.

### DIFF
--- a/.github/actions/free-hosted-runner-disk-space/action.yml
+++ b/.github/actions/free-hosted-runner-disk-space/action.yml
@@ -38,7 +38,7 @@ inputs:
     default: ''
     required: false
   diagnose-top-offenders-enabled:
-    description: Runs diagnostic steps to list top disk space offenders before and after cleanup
+    description: Runs diagnostic steps to list top disk space offenders before and after cleanup, and reports the size of each folder before it is removed
     required: false
     default: 'false'
 
@@ -74,7 +74,7 @@ runs:
         if [[ ${{ inputs.remove-dotnet }} == 'true' && -d $DOTNET_ROOT ]]; then
           echo "Removing .NET runtime and libraries..."
           start_time=$(date +%s)
-          du -sh $DOTNET_ROOT
+          [[ ${{ inputs.diagnose-top-offenders-enabled }} == 'true' ]] && du -sh $DOTNET_ROOT
           sudo rm -rf $DOTNET_ROOT
           end_time=$(date +%s)
           echo "Removed in $((end_time - start_time)) seconds"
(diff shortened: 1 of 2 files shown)


.github/actions/free-hosted-runner-disk-space/action.yml
```diff
@@ -71,107 +71,59 @@ runs:
...
+          fi
+        }
+
+        [[ ${{ inputs.remove-dotnet }} == 'true' ]] && remove_in_background ".NET runtime and libraries" "$DOTNET_ROOT"
+        [[ ${{ inputs.remove-android }} == 'true' ]] && remove_in_background "Android SDKs and Tools" "$ANDROID_ROOT"
+        [[ ${{ inputs.remove-haskell }} == 'true' ]] && remove_in_background "GHC (Haskell) artifacts" "$HASKELL_ROOT"
+        [[ ${{ inputs.remove-codeql }} == 'true' ]] && remove_in_background "CodeQL Action Bundles" "$CODEQL_ROOT"
+        [[ ${{ inputs.remove-swift }} == 'true' ]] && remove_in_background "Swift toolchain and libraries" "$SWIFT_ROOT"
+        [[ ${{ inputs.remove-powershell }} == 'true' ]] && remove_in_background "PowerShell Core installation" "$POWERSHELL_ROOT"
+
```
gionn:
```suggestion
```

pmacius (approved):

PR #1529, opened 2026-09-01
AAE-47554 Add shellcheck to actionlint pre-commit hook

andrphilippi:
### Checklist

- Jira Reference (also in PR title): [AAE-47554](https://hyland.atlassian.net/browse/AAE-47554)
  - [x] Patch (bugfix)
  - [ ] Minor (new feature)

### Description

- Issue Link: https://hyland.atlassian.net/browse/AAE-47554

Follow-up on the actionlint rollout. This repo already ran `actionlint` at the
latest `v1.7.12` with the `*.lock.yml` exclude, but it was missing the
`go-shellcheck` additional dependency, so inline `run:` scripts in workflows
were never actually shellchecked.

```yaml
additional_dependencies:
  - "github.com/wasilibs/go-shellcheck/cmd/shellcheck@v0.11.1"
```

`go-shellcheck` is a Wasm build, so contributors get shellcheck coverage
without needing a system `shellcheck` binary.

### Result

No lint fixes were required: `pre-commit run actionlint --all-files` passes
across all workflows with shellcheck enabled. Verified the integration is
genuinely active by temporarily linting a workflow with an unquoted variable,
which correctly reported `SC2086`.

Note that the upstream hook only targets `^\.github/workflows/`, so shell
embedded in composite `.github/actions/**/action.yml` files is still outside
its scope.

Made with [Cursor](https://cursor.com)

[AAE-47554]: https://hyland.atlassian.net/browse/AAE-47554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,6 +36,8 @@ repos:
       - id: actionlint
         args: [-ignore, 'unexpected key "queue" for "concurrency" section. expected one of "cancel-in-progress", "group"']
         exclude: ^(\.github/tests/|\.github/workflows/.*\.lock\.yml$)
+        additional_dependencies:
+          - "github.com/wasilibs/go-shellcheck/cmd/shellcheck@v0.11.1"
 
   - repo: https://github.com/hyland/github-actions-ensure-sha-pinned-actions
     rev: v1.2.0


.pre-commit-config.yaml
```diff
@@ -36,6 +36,8 @@ repos:
       - id: actionlint
         args: [-ignore, 'unexpected key "queue" for "concurrency" section. expected one of "cancel-in-progress", "group"']
         exclude: ^(\.github/tests/|\.github/workflows/.*\.lock\.yml$)
+        additional_dependencies:
+          - "github.com/wasilibs/go-shellcheck/cmd/shellcheck@v0.11.1"
```
gionn, line 40: what's the actual issue you are trying to solve?

I prefer we do not add a dependency maintained by a random guy with 11 stars on github

gionn (changes requested):

.pre-commit-config.yaml
JWollweber: The origin of this change comes from me. I did this on another PR - this here is just a roll out.

IIRC I've done this because the tool is pre-installed on GitHub runners and actionlint automatically uses it if it is installed on the machine. If you are not enforcing it in the pre commit config this means running pre-commit locally behaves different than on the CI. But let me double check as your concern is valid!

gionn: did a quick check and indeed shellcheck is already available on the CI:
https://github.com/Alfresco/alfresco-build-tools/pull/1531

so this is just a concern for developer machines which cannot install shellcheck easily (windows only and not using WSL?), IMHO not worth the risk of introducing it everywhere

JWollweber: ShellCheck is pre-installed on GitHub Ubuntu runners - e.g. look here: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2604-Readme.md

This one is probably based on the original from here https://github.com/koalaman/shellcheck
The problem is that it cannot be installed in pre-commit. I agree we should not add wasilibs version and require devs to install the tool manually if they want the same results as the CI. @andrphilippi can you remove the dependency from your PRs? Will we need a follow-up ticket to remove it from other repos?

andrphilippi:
See comment we decided not to introduce dependency to shellcheck. As this is the only change in this PR we can close it

.pre-commit-config.yaml
atchertchian: As it's auto-detecting and failing unexpectedly on CI, it could also be disabled explitcitly here with:
```
        args:
          - -shellcheck=
```

PR #1561, opened 2026-09-21
OPSEXP-4396 Migrate all internal action references to $/ self-repository syntax

gionn:
### Checklist

- Jira Reference (also in PR title): OPSEXP-4396
  - [ ] Patch (bugfix)
  - [ ] Minor (new feature)

### Description

This finishes the OPSEXP-4396 migration started by the canary (#1560): every remaining internal `uses:` reference, across all workflows and composite actions, now uses the `$/` self-repository syntax instead of an `Alfresco/alfresco-build-tools@<sha>` pin. The four folded-scalar refs in dbp-charts/verify-helm/action.yml collapse back to single lines, since there's no longer a 40-char SHA forcing them to wrap.

release.sh drops the two-pass SHA-pinning dance: the release-candidate commit, the eventual-consistency polling loop, and both sed passes are gone, since `$/` resolves against whatever commit is running without any of that. It now only bumps the version tags used in docs/ and per-action README/SKILL examples, and while widening that sed I fixed two examples that had already drifted out of sync (flux-operator-bootstrap, pimp-my-repo).

release.yml also drops the workflows:write permission request on the GitHub App token and switches from the org's ADMIN app to its CONTRIB app. release.sh no longer touches .github/workflows/ (only docs/ and per-action README/SKILL examples), so workflows:write is unneeded, and CONTRIB is already the app used for release workflows in other repos like alfresco-azure-connector. node_version is dropped too, since it only existed for the old release-candidate commit's npx step, which is gone.

check_action_nesting.py and its whole test tree are removed. The only thing it enforced was a max nesting depth of 3, and that number came entirely from how many release cycles the old two-pass scheme needed to converge, a problem that no longer exists. The Actions runner enforces its own nesting limit regardless. Its slot in pre-commit is replaced by a one-line check that no `$/` reference carries an `@ref` suffix, since that's invalid.

The `test/local-actions` label mechanism in test.yml is also removed: with `$/`, every PR tests its own nested action changes automatically, so the opt-in sed-to-local-path hack doesn't do anything useful anymore. The `test/local-actions` label itself stays in GitHub, since PR #1522 (unrelated, still open) currently carries it. The SHA-pin allowlist for `Alfresco/alfresco-build-tools/*` is dropped from both the pre-commit hook and CI, since it only matched owner/repo-style refs and none of those exist in this repo anymore.

docs/README.md gets a short Requirements note (runner 2.336.0+) and a line clarifying that this repo's internal `$/` usage has nothing to do with how consumers should pin their own references to us. copilot-instructions.md's internal-reference guidance is rewritten to match.

This ships as a major version: `$/` requires GitHub Actions runner 2.336.0 or newer, and any self-hosted runner fleet stuck below that breaks at runtime with no visible change on the consumer's side, since their `uses:` line doesn't change, only what it resolves to.

The canary (#1560) has landed and released as v18.29.0, and cross-repo resolution is confirmed: bumping acs-deployment's setup-helm-docs and setup-kind to that tag showed their internal `$/` refs resolving against alfresco-build-tools@v18.29.0, not against acs-deployment (see acs-deployment#1587).


docs/README.md
```diff
@@ -466,6 +467,15 @@ starts using [Immutable releases](https://docs.github.com/en/code-security/conce
...
 available via pre-commit as well.
 
+This repo composes its own actions internally using GitHub's `$/` self-repository syntax rather than pinning to
+itself, so the guidance above is about the actions and workflows *you* reference from `alfresco-build-tools`, not
+about how this repo's actions call each other.
+
+### Requirements
+
+Actions and reusable workflows in this repository require GitHub Actions runner **2.336.0 or newer**. GitHub-hosted
+runners already meet this; only older self-hosted runners may need an update.
```
gionn: remove

pmacius (approved):
🚀

dsibilio (approved):

docs/README.md
```diff
@@ -466,6 +466,10 @@ starts using [Immutable releases](https://docs.github.com/en/code-security/conce
 See [github-actions-ensure-sha-pinned-actions](https://github.com/hyland/github-actions-ensure-sha-pinned-actions),
 available via pre-commit as well.
 
+This repo composes its own actions internally using GitHub's `$/` self-repository syntax rather than pinning to
+itself, so the guidance above is about the actions and workflows *you* reference from `alfresco-build-tools`, not
+about how this repo's actions call each other.
+
```
mkrbr:
```suggestion
```

Not sure if we need to really add it into a readme... :D I feel like this assumes people are dumb.

